PR #9, opened 2017-08-18
Validate individual packages

alexkilgour:
Add command line argument to allow validation of a single package instead of always running on all the packages

```
// validate all packages
$ npm run validate

// validate only fec-package
$ npm run validate fec-package
```


scripts/validate.js
```diff
@@ -14,29 +14,48 @@ const getLicense = require('./_get-license');
...
-							});
-					}).catch(err => {
-						exitScript.displayErr(err);
-					});
-			}).catch(err => {
-				exitScript.displayErr(err);
-			});
-	}).then(() => i >= lastPackagePathIndex || pathLoop(i + 1));
-})(0);
+if (filteredPaths.length > 0) {
```
jpw: could we reduce this flying V of JS please? like bailing early on this if condition?

scripts/validate.js
```diff
@@ -14,29 +14,48 @@ const getLicense = require('./_get-license');
...
+	(function pathLoop(i) {
```
jpw: do we need IIFE's in node?

scripts/validate.js
```diff
@@ -14,29 +14,48 @@ const getLicense = require('./_get-license');
...
 const exitScript = require('./_exit-script');
 
 const allPackagePaths = getPackages(path.resolve(__dirname, '../packages'));
-const lastPackagePathIndex = allPackagePaths.length - 1;
 const globalLicense = getLicense('../package.json');
+const lastArgument = process.argv[process.argv.length - 1];
+const filteredPaths = filterPaths(lastArgument);
+const lastPackagePathIndex = filteredPaths.length - 1;
+
+// Filter the package list based on CLI argument
```
jpw: improve function name, rm comment?

alexkilgour:
col, thanks. will take a look at those

scripts/validate.js
alexkilgour: It's not needed in node, but i'm using the IIFE syntax to pass the `0` starting value

jpw: is the version of node you are targetting compatible with ES6 default parameters?

alexkilgour: yes, but you've still got to call the function right? Anyway, i'm having to rewrite it to get over the other issue so this is going anyway 😆

scripts/validate.js
alexkilgour: 👍

PR #10, opened 2017-11-03
Add more editorconfig defaults ported from shunter

josebolos:
This sets defaults for json, markdown and yml files.


alexkilgour:
👍

PR #18, opened 2018-01-19
Convert the package management into it's own package

alexkilgour:
Abstract out the scripts for creating, validating and publishing packages into it's own package. Then use that package on this repo. So Meta.

This will allow the same creation and validation scripts to be used on brand level repositories, and with a bit of extra work later, the publication step as well.

inception-gif-3


nickcall:
So many files.

Should this file run the other way latest at top ?
packages/global-package-manager/HISTORY.md

alexkilgour:
yep, you're right, it should!

PR #29, opened 2018-03-06
Switch from XO to ESlint

alexkilgour:
Use the new SpringerNature ESlint config instead of XO


.eslintignore
```diff
@@ -0,0 +1,2 @@
+**/__tests__/*
```
saramorueta: one question here! if you add the `__tests__` dir to be ignored, why to include `eslint-plugin-jest` as a dependency in `package.json`?

alexkilgour: I was just adding the ability to do it (to test it actually worked!) then i am going to lint that directory in a different PR

PR #30, opened 2018-05-17
Convert global-box to a component

alexkilgour:
As we are removing the pattern level, this converts this package to be a component.
Also adds `!default` to the settings to allow for inheritance of defaults.


davidpauljunior:
Wait what? No `!defaults`?

jpw:
Alex and I had a big chat about this and drew lots of things on a whiteboard, cant recall where we left the `!default` stuff though. We all need to have a chat about it 😄 

Also, I thought we were calling these packages not components?

PR #31, opened 2018-05-17
initial commit for copypaste component

o-t-w:


packages/global-copypaster/scss/10-settings/_global-copypaster.scss
```diff
@@ -0,0 +1,7 @@
+/**
+ * @springernature/global-box
```
davidpauljunior: global-copypaster?

o-t-w: Oops. This was copy and pasted. nicely spotted!

o-t-w:
Could somebody less clueless about testing than me tell me what sort of tests they would write for this component. Testing is a big black hole in my knowledge base 😭

alexkilgour:
Javascript should follow the guidelines from the playbook
https://github.com/springernature/frontend-playbook/blob/master/javascript/house-style.md#client-side-javascript-architecture

zanonnicola:
Using Jest you could just starting out by asserting that it actually copy the text. Something like this:

```js
import copyText from './copyText';

test('component copies the text', () => {

  document.body.innerHTML =
    `<button data-component="copypaste-button"] value="copy"/>
      <div [data-component="copypaste-text"]>My text</div>`;

  // Emulate a click on our button
  const copyButton = document.querySelector('[data-component="copypaste-button"]');
  copyButton.click();

  const {baseNode} = window.getSelection();
  // Assert that the fetchCurrentUser function was called, and that the
  // #username span's inner text was updated as we'd expect it to.
  expect(copyText).toBeCalled();
  expect(baseNode.nodeValue).toEqual('My text');
});
```

I haven't tried outthis code but it should point in the right direction at least

o-t-w:
@zanonnicola One thing I don't understand about testing modules:
If I want to test a function, it needs to be exported. Does this mean I have to export things I wouldn't otherwise export just so that I can test them with jest? The whole point of modules is keeping things out of the global scope.

zanonnicola:
@o-t-w usually you don't test private functions. Only public APIs. Even if JS does not have private methods per se they are not available in the current scope.

Unless you want to expose them only for testing and post-process your code I'd say leave them out. This is my view at least. I don't test implementation details.

This is a great workshop that covers modern JS testing: https://www.youtube.com/watch?v=VQZx1Z3sW0E&list=PLV5CVI1eNcJgNqzNwcs4UKrlJdhfDjshf
Hope it helps

danyalaytekin:
@o-t-w It does seem like some private functions have such clear interfaces and outputs that they are almost begging to be tested, and there are long discussions about whether this is appropriate on Stack Overflow, e.g. https://stackoverflow.com/questions/9122708/unit-testing-private-methods-in-c-sharp.

It can sometimes be a signal that, from a design point of view, those private functions should actually be extracted into their own module, even if testing were not a concern.  This excuse for extraction can usually be deployed without being questioned even when the extraction isn't appropriate.

packages/global-copypaster/view/index.html
```diff
@@ -0,0 +1,16 @@
+<div class="c-copypaste">
+	  <div class="c-copypaste__inner">
+		<span tabindex="-1" class="c-copypaste__text-to-copy" contenteditable="true" data-component="copypaste-text"></span>
+		<button aria-label="copy link to clipboard" class="c-copypaste__button" data-component="copypaste-button">Copy to clipboard <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
hollsk: The ARIA label isn't necessary as the button text itself ("copy to clipboard") should be sufficient.

I do think the "copy to clipboard" text can be made more meaningful by having something like "copy sharing link to clipboard".

packages/global-copypaster/view/index.html
```diff
@@ -0,0 +1,16 @@
+<div class="c-copypaste">
+	  <div class="c-copypaste__inner">
+		<span tabindex="-1" class="c-copypaste__text-to-copy" contenteditable="true" data-component="copypaste-text"></span>
```
hollsk: the contents of the contenteditable area are read out first, so the fact that there's an interaction isn't that obvious for screenreaders. ARIA describedby is a good way of fixing that, but i wasn't able to get it to work, and I think it might be because the focusing behaviour interrupts the screenreader (but not sure!). 

I think the easiest way to work around it might be by just moving the button in front of the contenteditable area in the source order, so it gets read out first. You don't necessarily have to change the styling so the button _looks_ like it's first visually though, since for keyboard users the whole widget gets selected when they tab into it.

hollsk: Voiceover announces this for the contenteditable area: "edit text, this is the text to copy, insertion at beginning of text" :thumbsdown:

But if you add `role="textbox"` and `aria-readonly="true"` to the contenteditable span, you'll get this instead: "this is the text to copy, clickable, text" :thumbsup:

packages/global-copypaster/view/index.html
```diff
@@ -0,0 +1,16 @@
...
+		<button aria-label="copy link to clipboard" class="c-copypaste__button" data-component="copypaste-button">Copy to clipboard <?xml version="1.0" encoding="UTF-8" standalone="no"?>
+		  <svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+			  <title>iconography/status-complete</title>
```
hollsk: "iconography/status-complete" isn't a very meaningful title for screenreaders :grimacing: Would be better if it's something like "copy successful" or something that describes what the image means

hollsk: The SVG title is announced by screenreaders before the interaction takes place. You can stop it happening by setting `aria-hidden="true"` on it initially, but then you need to find a way to announce the status change. 

It might be better to move the svg out of the button and put it inside an ARIA live region instead - something like `<div role="status" aria-live="polite"><svg blahblah></svg></div>`. Then you can set `aria-hidden="false"` at the right moment, and that'll update the status for screenreaders. 

But you should do that _after_ you've set the `.focus()` in the JS so the screenreader doesn't get interrupted by the focusing behaviour.

o-t-w: Which screenreader are you using? Using voiceover on safari and chrome the svg is not announced.

packages/global-copypaster/view/index.html
o-t-w: I was not aware of `aria-readonly="true"`. Exactly what I needed!

hollsk: Yeah, I think that's reasonable. If you're thinking of adding aria-hidden to the contenteditable area then it would do away with the need for the textbox role and gubbins too. 

I'd reword the text in the paragraph above to remove the "following" part, so SR users don't go looking for a non-existent link!

o-t-w:
I'm still confused by testing 😞 
The documentation says 

> Jest ships with jsdom which simulates a DOM environment as if you were in the browser. This means that every DOM API that we call can be observed in the same way it would be observed in a browser!

Yet Jest doesn't seem to understand `document.createRange` or `window.getSelection` so I'm having to mock so many things that what jest ends up testing bears barely any relation to my actual code.

zanonnicola:
@o-t-w can you share your jest config?

o-t-w:
@zanonnicola Rather than using a `jest.config.js` file i'm configuring in `package.json`. Pretty minimal:
```
  "jest": {
	"setupFiles": [
		"./jeststuff.js"
	]
}

contents of that file:
```
document.body.innerHTML =
`document.body.innerHTML =
`<div class="c-copypaste">
<div class="c-copypaste__inner">
  <span tabindex="-1" role="textbox" aria-readonly="true" class="c-copypaste__text-to-copy" contenteditable="true" data-component="copypaste-text"></span>
  <button aria-label="copy link to clipboard" class="c-copypaste__button" data-component="copypaste-button">Copy to clipboard </button><?xml version="1.0" encoding="UTF-8" standalone="no"?>
  <svg class="success-icon" width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
	  <title>copy successful</title>
	  <defs></defs>
	  <g id="iconography/status-complete" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
		  <g id="Group-10" transform="translate(2.000000, 2.000000)">
			  <circle id="Oval-2" fill="#3D993D" cx="10" cy="10" r="10"></circle>
			  <path d="M8.07715491,12.0625 C8.50813743,11.2113095 8.87281495,10.4255952 9.30379747,9.67261905 C9.66847498,9.01785714 10.1657625,8.19940476 10.729355,7.47916667 C11.2929476,6.69345238 11.7239301,6.13690476 12.0886076,5.77678571 C12.4532851,5.41666667 12.7185051,5.1547619 12.8842676,5.12202381 C13.0500301,5.08928571 13.3152502,4.92559524 13.7462327,4.89285714 C14.1109102,4.82738095 14.4092827,4.76190476 14.6745027,4.76190476 C14.8734177,4.76190476 14.9728752,4.82738095 15.0391802,4.89285714 C15.1717902,5.02380952 15.2380952,5.12202381 15.2380952,5.25297619 C15.2380952,5.38392857 15.2380952,5.48214286 15.1717902,5.54761905 C15.1054852,5.61309524 14.9728752,5.77678571 14.8071127,5.9077381 C13.9451477,6.69345238 13.0168776,7.87202381 12.0223026,9.3125 C11.0940325,10.7857143 10.2983725,12.2261905 9.73477999,13.5684524 C9.50271248,14.125 9.37010247,14.485119 9.23749247,14.6160714 C9.17118746,14.9107143 9.03857746,15.0416667 8.87281495,15.1071429 C8.67389994,15.2380952 8.37552743,15.2380952 7.8782399,15.2380952 C7.51356239,15.2380952 7.31464738,15.2380952 7.21518987,15.172619 C7.08257987,15.1071429 6.98312236,15.0416667 6.91681736,14.9761905 C6.85051236,14.8452381 6.68474985,14.6815476 6.48583484,14.3541667 C6.18746233,13.8630952 5.75647981,13.3720238 5.25919228,12.8154762 C4.96081977,12.4553571 4.76190476,12.1934524 4.76190476,11.9642857 C4.76190476,11.6696429 4.96081977,11.4077381 5.25919228,11.1130952 C5.6238698,10.8839286 5.92224231,10.7529762 6.18746233,10.7529762 C6.48583484,10.7529762 6.81735986,10.8839286 7.11573237,11.047619 C7.48040989,11.3422619 7.7787824,11.6369048 8.07715491,12.0625 Z" id="Shape" fill="#FFFFFF"></path>
		  </g>
	  </g>
  </svg>
</div>
</div>
`;
```

zanonnicola:
Yeah `document.createRange` is not supported in JSDOM (looks like) :( so you have to mock it.

jpw:
Sorry for the late feedback on this but;
- should not be bound to anything in the DOM as part of the component. Suggest passing in selector(s) and having an init function, and losing the button
- should be in ES6 & transpiled, so no lets.

o-t-w:
@jpw Did another commit.

alexkilgour:
Sorry, just coming back to this again. I think that this component should just be the JS and remove all the CSS. The component is the functionality and the styling is irrelevant here.

o-t-w:
@alexkilgour The copy paste function isn't very complicated. If time and thought has gone into interaction/UX, accessibility, etc (which are tied up with styling) its not irrelevant.

PR #36, opened 2018-06-08
Colors

o-t-w:


packages/global-colors/scss/10-settings/_global-colors.scss
```diff
@@ -3,22 +3,6 @@ $colors: (
 	'text-secondary': #666,
```
alexkilgour: I think this is still too much. This is still making assumptions about colors that we don't need. All you need here is an example of _how_ to use this.
```
$colors: (
	'primary': #000,
	'secondary': #fff,
);
```

PR #39, opened 2018-06-13
Colorhistory

o-t-w:
change $brand-colors to $colors


davidpauljunior:
Not really related to this PR, but should the map be `$global-colors`?

PR #40, opened 2018-06-21
Expandable box

o-t-w:


alexkilgour:
Does this have an actual use case for one of our projects?
Also, you don't have to have the `scss` and `view` folders if this is a js component

o-t-w:
Its used on Nature. I just wanted a js thing to use in the Vertical Slice.

packages/global-expandable-box/js/index.js
```diff
@@ -0,0 +1,55 @@
...
+	} else {
+		open(box, closetext);
+		box.isOpen = true;
+	}
+}
+
+function init(boxes, height = '95px', closetext = 'Show less', opentext = 'Show more') {
+	boxes.forEach(box => {
+		let buttonEtc = '<div><div></div><div><button type="button" aria-expanded="false"><span>Show more</span></button></div></div>';
+		box.insertAdjacentHTML('besforeEnd', buttonEtc);
```
joeyciechanowicz: `'besforeEnd' => 'beforeend'`

alexkilgour:
We should have full details of how to use this in the `README.md`

alexkilgour:
Your build is failing. You can use `npm run validate` to check issues locally

packages/global-expandable-box/js/index.js
```diff
@@ -0,0 +1,55 @@
...
+		box.isOpen = false;
+	} else {
+		open(box, closetext);
+		box.isOpen = true;
+	}
+}
+
+function init(boxes, height = '95px', closetext = 'Show less', opentext = 'Show more') {
+	boxes.forEach(box => {
+		let buttonEtc = '<div><div></div><div><button type="button" aria-expanded="false"><span>Show more</span></button></div></div>';
```
alexkilgour: what's with all the `divs`?

alexkilgour: also, should `Show more` come from the init param rather then being hardcoded into the button here?

PR #43, opened 2018-06-21
bump version number in package.json

o-t-w:


alexkilgour:
I think we should start with version `0.1.0` here, as per the semver specs
https://semver.org/#spec-item-4

(this is something I have got wrong before in previous components)

PR #45, opened 2018-06-22
Just jesting

alexkilgour:
Jest had a bug where you could not use `--silent` and `--verbose` together, which led to using a very convoluted process to mock console output (also, I didn't understand Jest very well at this stage!)

This bug has [now been fixed](https://github.com/facebook/jest/blob/master/CHANGELOG.md#jest-2120) 🎉 which means I can remove a bunch of code.

This change:
- removes the mock file for logging
- removes `showOutput.simpleLogToConsole` as we can now just use `console.log`
- simplify `showOutput.logToConsole` to `showOutput.log`


josebolos:
Silly question, how an app can be --silent and --verbose at the same time? 🤔

alexkilgour:
because `verbose` is meant to display the details of each individual test, but `silent` is meant to suppress console output, which means they can be used together
https://github.com/facebook/jest/issues/4458

PR #50, opened 2018-06-25
Publishbug

alexkilgour:
not ready yet. testing.


packages/global-package-manager/js/_validate.js
```diff
@@ -54,15 +58,37 @@ function allPackagePathLoop(i) {
...
+							config.packagesDirectory,
+							config.changelog,
+							!validatePublication
+						);
+					})
+					.catch(err => {
+						exitScript.displayErr(err);
+					});
+			}
+		} else {
```
jpw: can we avoid this `else` too?

packages/global-package-manager/js/_validate.js
```diff
@@ -54,15 +58,37 @@ function allPackagePathLoop(i) {
...
+			allPackagePathLoop(i + 1);
+		}
+	});
```
jpw: am finding this is a bit of a headbuster and the code is getting nested, can we break out the bit that checks one package into a function, and have the loop call that?

packages/global-package-manager/js/_validate.js
```diff
@@ -54,15 +58,37 @@ function allPackagePathLoop(i) {
...
-	}).then(() => i >= lastPackagePathIndex || allPackagePathLoop(i + 1));
+	}).then(() => {
+		if (i >= lastPackagePathIndex) {
+			if (validatePublication) {
+				configGenerator(
+					'package-manager.json',
+					path.resolve(process.cwd(), 'package.json')
+				)
+					.then(config => {
+						publishPackages(
```
jpw: is this publishing in a script called validate? its confusing what this code is actually doing

packages/global-package-manager/js/_validate.js
```diff
@@ -54,15 +58,37 @@ function allPackagePathLoop(i) {
...
+					});
+			}
+		} else {
+			allPackagePathLoop(i + 1);
+		}
+	});
 }
 
-module.exports = (packageJsonPath, configJson, arg) => {
+module.exports = (packageJsonPath, configJson, argP, argN) => {
```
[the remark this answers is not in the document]
alexkilgour: good point. hopefully this is better now

packages/global-package-manager/js/_validate.js
alexkilgour: yes but it's a bit tricky here, but this won't look anything like this once the work here is finsihed

packages/global-package-manager/js/_validate.js
alexkilgour: yep, again, this won't be needed once i can properly split the validate and publish steps

packages/global-package-manager/js/_validate.js
alexkilgour: this won't be needed once I can properly split the validate and publish steps

PR #52, opened 2018-06-25
Paths

o-t-w:
Including node_modules manually in a path seems to clash with using path resolution. Its nicer to not have to type it every time :)
e.g. npm script
```
"sass": "node-sass --include-path='./node_modules' --output-style compressed -o public/css src/sass/style.scss --sourcemap"
```


davidpauljunior:
Should this be added into the README or global-toolkit docs somewhere?

jpw:
Good point David! Ollie is just finishing off the vertical slice and then we will document the findings.

PR #59, opened 2018-07-04
Add more detail to readme

alexkilgour:
- Added missing info on how Lerna is used
- More examples on configuring with TravisCI


o-t-w:
Where are `changelog` and `sn-package-publish`? I can't find them in the project.

alexkilgour:
The changelog is set in the config - https://github.com/springernature/frontend-global-toolkit/blob/master/packages/global-package-manager/README.md#default-configuration

alexkilgour:
`sn-package-publish` is defined in the package.json - https://github.com/springernature/frontend-global-toolkit/blob/master/packages/global-package-manager/package.json#L11

and described here - https://github.com/springernature/frontend-global-toolkit/blob/d44ecfafa949214362ba35255e0d35cd09694a8d/packages/global-package-manager/README.md#install

PR #60, opened 2018-07-05
probably terrible code

o-t-w:


packages/global-package-manager/js/config/default.json
```diff
@@ -1,6 +1,7 @@
 {
 	"packagesDirectory": "./packages",
 	"changelog": "HISTORY.md",
+	"changelogrequired": "true",
```
danyalaytekin: Probably a terrible comment, but should this be `true`?

benjclark: What a fantastic comment. Good spot.

PR #61, opened 2018-07-11
Branding a global component

alexkilgour:
We want to allow a `global` component to be branded in a `brand` repository. The process for this is to create a new component in your brand repo, and specify a special kind of dependency on a component from the global toolkit. In the code this is known as `extending` (i.e. brand component X extends global component Y). Any files that exist in the global component that do not have a path that matches a file in the brand component, are merged into the brand component at the publish step. This new merged component is then published to NPM.

1. Add [setting](https://github.com/springernature/frontend-global-toolkit/blob/1e539133caa091062bddec85a86f81a599222134/packages/global-package-manager/js/config/default.json#L5) to allow or disallow branded components from within a repo (this is so that brand repos can extend but the global repo can be set up so as not to extend a component from itself)
2. A package that extends another package needs to have a key/value in it's package.json of the form `"extendsPackage": "global-name-of-component@version"` e.g. `"extendsPackage": "global-colors@1.0.0"`
3. When running the validation and publication steps we [check for this key](https://github.com/springernature/frontend-global-toolkit/blob/1e539133caa091062bddec85a86f81a599222134/packages/global-package-manager/js/_extend-package.js#L78) to see if the package is extending another package
4. At the validation step, if the check above returns true, validation is passed if the package we are trying to extend [can be found with the correct version](https://github.com/springernature/frontend-global-toolkit/blob/1e539133caa091062bddec85a86f81a599222134/packages/global-package-manager/js/_extend-package.js#L69)
5. At the publication step, if the check above returns true, the package is published after the successful completion of the following:
    - [Get list of files](https://github.com/springernature/frontend-global-toolkit/blob/1e539133caa091062bddec85a86f81a599222134/packages/global-package-manager/js/_get-extended-file-list.js) from the package being extended (the dependency)
    - [Merge dependency files](https://github.com/springernature/frontend-global-toolkit/blob/1e539133caa091062bddec85a86f81a599222134/packages/global-package-manager/js/_merge-extended-package.js) with the package if they do not already exist

FYI: I think before releasing a new version of this package it would be best to move it out of here into it's own repo


packages/global-package-manager/js/_check-extended-package.js
```diff
@@ -0,0 +1,32 @@
...
+							type: 'success',
+							description: 'validating',
+							message: packageToExtend
+						}]);
+						resolve();
+					}).catch(err => {
+						reject(err);
+					});
+			}).catch(() => {
+				resolve();
```
o-t-w: Why do you want this catch block to resolve()?

alexkilgour: because this is checking if the current package is an extended package. if it isn't we still carry on, we just don't validate it as an extended package

o-t-w: Oh. The amound of resolve()'s and catch()'s and reject()'s is a little confusing. Would this work?: 
```
function validateAndLogExtendedPackage(config, packagePath) {
    return extend.validate(packagePath, config.allowExtends, config.scope)
        .then(packageToExtend => {
            showOutput.log([{
                type: 'success',
                description: 'validating',
                message: packageToExtend
            }]);
        });
}

function checkExtendedPackage(config, packagePath) {
    return extend.check(packagePath)
        .then(
            () => validateAndLogExtendedPackage(config, packagePath),
            () => null
        );
}
```

packages/global-package-manager/js/_check-package-structure.js
```diff
@@ -0,0 +1,176 @@
...
+		file: {messaging: [], valid: true},
+		topLevel: {messaging: [], valid: true}
+	};
+
+	results = JSON.parse(JSON.stringify(blankResults));
+	requiredFiles = JSON.parse(JSON.stringify(config.required));
+
+	return new Promise((resolve, reject) => {
+		checkPackageStructure(results, packagePath, options)
+			.then(resolve)
```
o-t-w: should it be `.then(resolve())` ?

alexkilgour: nothing is being passed so you don't need the `()`, this is enforced by the linter

packages/global-package-manager/js/_check-extended-package.js
alexkilgour: Not sure this ever resolves successfully to carry on. I also find `() => null` to be more confusing. The whole package needs to be looked at because at the start I never anticipated it getting this large so I'm not sure the overall pattern works now. But this follows a very definitive pattern throughout the package which I don't want to change unless everything changes with it. I can add a comment to make it clear what is going on here.

alexkilgour:
I have simplified the code here by removing the `_check-extended-package.js` and `_get-extended-package.js` functions, but with the cost of adding more complexity to `_extend-package.js`

PR #63, opened 2018-07-20
adds dropdown component

o-t-w:


packages/global-dropdown/js/index.js
```diff
@@ -0,0 +1,12 @@
+function init() {
```
zanonnicola: Is it worth to check for the existence of the DOM element?
Not sure if you decide to handle this outside the component!

packages/global-dropdown/package.json
```diff
@@ -0,0 +1,13 @@
...
+	"version": "0.1.0",
+	"license": "MIT",
+	"description": "dropdown component",
+	"keywords": [
+	  "css",
+	  "js",
+	  "component"
+	],
+	"author": "Oliver Williams"
+  }
```
jpw: indentation

packages/global-dropdown/js/index.js
```diff
@@ -0,0 +1,12 @@
...
+	var dropitbutton = [...document.querySelectorAll('[data-dropdown]')];
+
+	dropitbutton.forEach(button => {
+		button.addEventListener('click', function () {
+			this.nextElementSibling.classList.toggle('display-none');
+			this.querySelector('svg').classList.toggle('rotate');
```
zanonnicola: IE11 does not support `toggle` for the SVG element: https://caniuse.com/#feat=classlist :(

packages/global-dropdown/js/index.js
```diff
@@ -0,0 +1,12 @@
+function init() {
+	var dropitbutton = [...document.querySelectorAll('[data-dropdown]')];
```
jpw: 1. `dropitbutton` is a weird variable name, and not camel-cased
2. we have a spread operator (ES6) but use of `var` not `const`

packages/global-dropdown/js/index.js
jpw: consider [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) in this code

packages/global-dropdown/js/index.js
nasiamakryg: I used this polyfill and everything worked like charm: https://github.com/springernature/snap/pull/237/commits/d4b24f54cac956d3470bcef467ddef16f325a4be

o-t-w: I think CSS transforms don't work at all on svgs in IE. The dropdown continuing to point in the same direction perhaps isn't a particularly bad user experience? What do people think?

packages/global-dropdown/scss/50-components/_global-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+	color: $global-dropdown--color;
+	display: flex;
+}
+
+.c-dropdown__button:hover {
+	background-color: white;
+	color: $global-dropdown--hover-color;
+}
+
+.display-none {
```
nasiamakryg: if this is a util class then maybe you wanna write it as .u-display-none

packages/global-dropdown/scss/50-components/_global-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+}
+
+.c-dropdown__menu-button {
+	display: block;
+}
+
+.c-dropdown__menu {
+	margin-top: 10px;
+	position: absolute;
+	color: #666666;
```
saramorueta: isn't there any variable that suits the colour?

packages/global-dropdown/scss/50-components/_global-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+	background-color: white;
+	border: solid #C0C0C0 1px;
+	box-shadow: 0 0 5px rgb(200,200,200);
+}
+
+button {
+	min-width: 170px;
+	padding: 10px 10px;
```
saramorueta: Same question than above regarding the spacing :blush:

packages/global-dropdown/scss/50-components/_global-dropdown.scss
```diff
@@ -0,0 +1,95 @@
...
+
+.c-dropdown__button g {
+	fill: currentColor;
+}
+
+.c-dropdown__button:hover g {
+	fill: currentColor;
+}
+
+.rotate {
```
nasiamakryg: this css rule could be applied as a modifier class on the elem, for example .c-dropdown__button--rotated in order to follow our bemish conventions

packages/global-dropdown/js/index.js
o-t-w: In reply to 3: Maybe. What issue might people have with using `data-dropdown` in their markup?

packages/global-dropdown/scss/50-components/_global-dropdown.scss
o-t-w: Its a class that only does one thing and will be reusable elsewhere so I think it would be good to have it as a utility that can be reused elsewhere.

packages/global-dropdown/view/dropdown.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="c-dropdown">
```
saramorueta: just a reflection, some time ago I found this example from W3C that made me think if we should use a normal list and the power of WAI-ARIA
https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html

packages/global-dropdown/js/index.js
nasiamakryg: They do! 
this is the stackoverflow thread explaining the trick with classlist and svgs:
https://stackoverflow.com/questions/8098406/code-with-classlist-does-not-work-in-ie/8098430#8098430

screen shot 2018-07-20 at 11 46 50 am

packages/global-dropdown/scss/50-components/_global-dropdown.scss
o-t-w: It would be nice for this to be a variable. Its going to be the same colour across all sites, so I didn't want to put it in the settings file where the component specific variables live. I guess elsewhere in the codebase should have a variable for this, but it shouldn't be defined at a component level? Hopefully we can have consistent color variable names across sites for things that are consistent across brands.

packages/global-dropdown/scss/50-components/_global-dropdown.scss
nasiamakryg: I am wondering if it is a good idea then to define util classes inside components.. I would avoid it and generally use as less util classes as possible but this is a nice subj to discuss on the fed open space ;)

packages/global-dropdown/view/dropdown.hbs
jpw: ^ that example does look nice!

packages/global-dropdown/js/index.js
o-t-w: Hey @zanonnicola. I'm never sure about this! This code doesn't error even if there is no element on the page with the relevant data attribute ([data-dropdown]). Could you show me how you would write it?

packages/global-dropdown/view/dropdown.hbs
hollsk: I can, but before I do, what's the intended _purpose_ of the component? What are its use cases?

packages/global-dropdown/js/index.js
o-t-w: If you use an arrow function as a callback in an event listener, `this` will be `undefined`.

packages/global-dropdown/scss/50-components/_global-dropdown.scss
o-t-w: I agree. It is challenging writing code without any real context of knowing what else will be in the codebase! I think we need to start looking at a small utility class "component"/library. @alexkilgour @jpw

packages/global-dropdown/view/dropdown.hbs
o-t-w: It is not meant to be a replacement for a <select> element. All sites will continue to use the native HTML element. It is not meant to submit data to a form. The main use case will be downloading files. 
screen shot 2018-07-23 at 11 42 49

hollsk: Aha! Ok, that's very helpful, thank you! 

I wouldn't use a listbox for this, but I do agree with Sara that it should use a list for the items inside the dropdown, so that screenreader users know how many options they have and when they get to the end of them. I also agree with Sara that it can use some aria to indicate that it's a popup. Have a look at the "submenus" example here: http://heydonworks.com/practical_aria_examples/#submenus

FWIW if it's not submitting to a form, then I'd probably also not use buttons either - I'd just use a links. Huge article on that here: https://css-tricks.com/building-good-download-button/

o-t-w: Thanks Holly. I had a read. I added `aria-haspopup="true"` to the button that reveals the dropdown and the div into a `ul`.

package-manager.json
```diff
@@ -14,7 +14,8 @@
     ],
     "view": [
       "html",
-      "dust",
+	  "dust",
```
alexkilgour: Just get rid of dust here. not sure anything is using it yet, and we seem to have settled on handlebars

packages/global-dropdown/view/dropdown.hbs
```diff
@@ -0,0 +1,8 @@
...
+	<ul class="c-dropdown__menu display-none">
+		<li class="c-dropdown__menu-button"><a download href="prongles.mp3">option 1</a></li>
+		<li class="c-dropdown__menu-button"><a download href="prongles.mp3">option 1</a></li>
+		<li class="c-dropdown__menu-button"><a download href="prongles.mp3">option 1</a></li>
+	</ul>
+</div>
```
alexkilgour: This is a dynamic template, you shouldn't be hardcoding text into here, e.g. `option 1`. You need to decide on a data structure for this.

Also, what is the word `download` for on the anchor link?

And I am assuming that the color of the SVG should be configurable, so this should be handled via CSS and a settings config.

packages/global-dropdown/view/dropdown.hbs
```diff
@@ -0,0 +1,8 @@
+<div class="c-dropdown">
+	<button aria-haspopup="true" class="c-dropdown__button" data-dropdown>dropdown <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="10" height="6" viewBox="0 0 10 6"><defs><path id="chevron-down-a" d="M12.5857864,13 L9.29289322,9.70710678 C8.90236893,9.31658249 8.90236893,8.68341751 9.29289322,8.29289322 C9.68341751,7.90236893 10.3165825,7.90236893 10.7071068,8.29289322 L14.7071068,12.2928932 C15.0976311,12.6834175 15.0976311,13.3165825 14.7071068,13.7071068 L10.7071068,17.7071068 C10.3165825,18.0976311 9.68341751,18.0976311 9.29289322,17.7071068 C8.90236893,17.3165825 8.90236893,16.6834175 9.29289322,16.2928932 L12.5857864,13 Z"/></defs><g fill-rule="evenodd" transform="translate(-7 -10)"><rect width="24" height="24" opacity="0"/><use fill-rule="nonzero" transform="rotate(90 12 13)" xlink:href="#chevron-down-a"/></g></svg></button>
```
hollsk: The SVG arrow is decorative, so you can add `role="presentation"` to it, to hide it from assistive technology :-)

packages/global-dropdown/view/dropdown.hbs
hollsk: nicely, thanks! 👍 

I've been having a further look at it (and at Heydon's example). Both are incomplete 😬The more I think about it, the more I'm coming to believe that this should be a _menu_, with the various menu roles attached. WAI-ARIA have an example here: https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html

The main things they have that are different is the addition of aria-expanded (where necessary), and focus management on interaction, as well as the appropriate roles. 

Heydon has also written a very, very long article on this subject - the pertinent bit for us is the "True Menus" section (but the whole article is good):  https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/ 

He's also provided a github link at the end to the entire component. 

Sorry! I've just turned this into a metric crapton of work 😰

hollsk: Also, just to add - I prefer the WAI-ARIA example to the heydon/giraud example because it uses the list. Basically it feels more progressively enhanced to me to use list semantics to start with, and to add the `role="none"` if the JS is available.

o-t-w: Should it have both aria-expanded and haspopup?

hollsk: Haha yeah, sorry for the huge article. It explains why I'm coming to believe the component should be a menu, though 😆 I still think the links should be links. The examples aren't exactly what you're doing here, so it's ok IMO. 

I don't know why the WAI-ARIA example removes the semantics on the list either, but it's not progressively-enhanced, so if somebody using AT is trying to use it without JS, they're SOL as it doesn't work at all. That said, yours has the same problem, doesn't it? 

r.e. aria-expanded and haspopup - yes. Here's another massive article SORRY LOL https://www.marcozehe.de/2010/02/10/easy-aria-tip-5-aria-expanded-and-aria-controls/ - it doesn't mention aria-haspopup (aria-controls only works in JAWS and is a bit ropey so I'm on the fence about whether that should be implemented or not - i don't know enough to be able to say). 

Let me get back to you on the menu items vs links thing, I need to do a bit of digging on that 😄

packages/global-dropdown/view/dropdown.hbs
hollsk: haha what, no, thats not what `role="presentation"` does, I've gone mad. 

Use `aria-hidden` instead 🤦‍♀️

packages/global-dropdown/view/dropdown.hbs
hollsk: Right, I've done some digging. The answer is, I don't know 😆 

Support for the `download` attribute in screenreaders doesn't exist, so screenreaders aren't getting any specific benefit from that right now, regardless of whether the items are announced as links or menu items. Perhaps the attribute will be supported later, as it's still quite new. If it were supported, I don't know what the interaction would be like, but right now the only difference is the SR saying "link" vs "menu item". 

In light of SRs not picking up the `download` attribute, I think it'd be worthwhile to add a hint, somehow. You _could_ use `aria-describedby` on each link to refer back to the button. In Voiceover, for the example screenshot you've posted above, that would mean the content would be read out, and then there'd be a short pause, and then "Export citation" would be read out after that. An alternative would be to add a visually-hidden `<span>Download</span>` at the start of the link, which might make the action more explicit 🤔Dunno, maybe. I'm handwaving a bit here. 

----

Back to the link vs menu item bit! The WAI-ARIA example has a bit of chat at the start of the demo:

>This example demonstrates the menu button design pattern for a button that displays a menu of link targets. The menu items are made from HTML links and maintain their HTML link behaviors. That is, activating a menuitem loads the link target, and the browser's link context menu and associated actions are available.

... so I infer from this that WAI themselves don't necessarily think it's a problem, so we're back to whether the component semantically is a menu or not. 

Going back to Heydon's article, he mentions near the start that you can tell the difference between a navigation menu and a true menu by looking at the trigger for opening - does it have a location of its own? If it does, then it's a navigation menu. But ours doesn't - ours is more like this example:

>Menu buttons which take the aria-haspopup="true" property do not behave like this. They are activated on click and have no other purpose than to reveal a secreted menu.

The "export citation" button doesn't _do_ anything on its own, other than to show the options available. 

But then again, maybe it _is_ a table of content. You could rewrite the whole component to be a flat heading of "Export citation" with a list underneath with all the content in, and it'd make sense semantically 🤔 

uuuuuurgh nobody ever said semantics was easy 😩

hollsk: WAI-ARIA's [`menu` specification](https://www.w3.org/TR/wai-aria-1.1/#menu) says:

>A menu is often a list of common actions or functions that the user can invoke. The menu role is appropriate when a list of menu items is presented in a manner similar to a menu on a desktop application.

Does this component count? Well, maybe. I have plenty of apps that offer an "Export..." option that lets me save a file to my hard drive. And as a user, _downloading_ a file feels different to _navigating_ to a file in the browser, even though the underlying semantics are pretty much the same. 

On the other hand, the screenshot has a lot of markup inside the `a` elements, doesn't it? It's not just "RIS", it's "RIS"+all that other stuff. That makes it look less like a menu, and more like, i dunno, some kind of random popup. (Btw, the `aria-haspopup` part will need to be removed if it's not a menu, ironically, lol)

o-t-w: I've added some aria labels. The HTML is just an example that people can change.

hollsk: Aria labels override the text of the links - so instead of hearing "option 1", "option 2", etc, people will hear "download file" for all of them and won't know what they're downloading 😉  

The advantage of aria-describedby is that it's supplementary - it's designed to associate an element with something else that's visible on the screen (e.g. the button content). The visually-hidden span method means that you can add a hidden hint for people who don't have access to the visuals.

packages/global-dropdown/js/index.js
```diff
@@ -0,0 +1,81 @@
...
+	}	else if (button.getAttribute('aria-expanded') === 'true') {
+		closeDropdown(button);
+	}
+}
+
+function init(selector = '[data-dropdown]') {
+	const dropdownButtons = [...document.querySelectorAll(selector)];
+
+	dropdownButtons.forEach(button => {
+		button.addEventListener('click', openOrCloseDropdown.bind(null, button));
```
hollsk: Coming together nicely! 

You need to listen for a keydown event here to detect the space and return keys for screenreaders, as they don't use the click event.  That WAI-ARIA example also detects the down arrow, for good measure ( https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/js/Menubutton2.js ).

Aside from that it's looking really good - it's easy to use and it looks robust. ✨

o-t-w: Because it is a button doesn't the browser automatically make the click event also work for both the space key and return key? Both the return button and the space button open the dropdown for me (at least in Safari).

hollsk: Doesn't work for me in Chrome + Voiceover 🙂

jpw:
Why is the config arg  "pointless" ?
Also is it ok to use `KeyboardEvent.key` ? https://caniuse.com/#feat=keyboardevent-key

o-t-w:
@jpw Is it foreseeable that people will be unable to insert the correct data-attributes inside their markup? If so, I will change it, but I think it makes the code slightly more opaque.

o-t-w:
@jpw I've changed it so that you can now pass in alternative selectors to the init function.

o-t-w:
@jpw I've now added code to use (deprecated) keyCode when event.key is not available.

packages/global-dropdown/view/dropdown.html
```diff
@@ -0,0 +1,8 @@
+<div class="c-dropdown">
+	<button aria-expanded="false" aria-haspopup="true" class="c-dropdown__button" data-dropdown>dropdown <svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="10" height="6" viewBox="0 0 10 6"><defs><path id="chevron-down-a" d="M12.5857864,13 L9.29289322,9.70710678 C8.90236893,9.31658249 8.90236893,8.68341751 9.29289322,8.29289322 C9.68341751,7.90236893 10.3165825,7.90236893 10.7071068,8.29289322 L14.7071068,12.2928932 C15.0976311,12.6834175 15.0976311,13.3165825 14.7071068,13.7071068 L10.7071068,17.7071068 C10.3165825,18.0976311 9.68341751,18.0976311 9.29289322,17.7071068 C8.90236893,17.3165825 8.90236893,16.6834175 9.29289322,16.2928932 L12.5857864,13 Z"/></defs><g fill-rule="evenodd" transform="translate(-7 -10)"><rect width="24" height="24" opacity="0"/><use fill-rule="nonzero" transform="rotate(90 12 13)" xlink:href="#chevron-down-a"/></g></svg></button>
+	<ul class="c-dropdown__menu u-display-none">
```
sonniesedge: How does this work if JS fails to load? Won't the list be invisible in that case?

packages/global-dropdown/scss/50-components/_global-dropdown.scss
```diff
@@ -0,0 +1,98 @@
...
+	width: 100%;
+}
+
+.c-dropdown__button {
+	border: solid $global-dropdown--border-color 1px;
+	background-color: $global-dropdown--background-color;
+	color: $global-dropdown--color;
+	display: flex;
+	min-width: 170px;
+	padding: 10px 10px;
```
sonniesedge: Wouldn't these (and other sizing variables) be better off as Sass variables? I could then reconfigure values as I use them. Right now I'd have to override these vales later in CSS, raising the specificity.

o-t-w: I think James Alegria was hoping the look would be exactly the same (except for fonts and colors) in all projects. Maybe that idea is not practical?

jpw:
Ollie and I just paired on this to rejig the JS and do some other tweaks. @sonniesedge your question is still open re. the CSS, cos I don't know the answer 😄 thanks for having a look!

Ollie and I intend to do some tests tomorrow, then if we can resolve the CSS question we might just be finished!

packages/global-dropdown/scss/50-components/_global-dropdown.scss
o-t-w: I think there is a desire to have all buttons across all sites have consistent padding, because why not?

o-t-w: Predicting what people will and won't want to customise is pretty challenging for component authors tbh. Is overwriting things by using _slightly_ more specificity the worst thing?

o-t-w: CSS variables can have default values. Can we do the same thing with Sass? I think it would be a good to have the same thing with our component variables so that people that want to use them are not forced to define every single variable if they cba.

o-t-w: How many & which other sizing variables are you thinking about? That would be a lot of configuration

o-t-w: Want to double check @alexkilgour Am I right in thinking there are no default values, so you have to specify a value for every Sass variable?

sonniesedge: > Is overwriting things by using slightly more specificity the worst thing?

Yes - I'm absolutely going to say yes to this. The more we can shift values into configuration, the better. As soon as you start a specificity war in the CSS layer it escalates out of control. Better to never ever start it by keep things at minimum-possible levels. 

As for consistency, yes I think that is absolutely a priority. But I think that hardcoding values with the assumption that nothing will _ever_ change would be a slightly naive one. In reality something will change - that's just life. No plan ever survives contact with the enemy. 😂 Better to let it be changed in a product by overriding a Sass variable, rather than forcing it to be overriddenn in CSS. 

Why not just specify these as variables? I feel like I'm missing an argument here (please tell me if I am):

In the component:
`$dropdown-min-width: 170px !default; `

'min-width: $dropdown-min-width;'

In a settings file added per project (optional): 

`$dropdown-min-width: 200px`;

o-t-w: My specificity comment was a throwaway comment which I have now deleted as I am happy to agree about that. 

@alexkilgour was inclined to not use `!default`. I can't remember his reasoning. Personally I think we should be using `!default`, particularly if we're having a lot of config options.

sonniesedge: @alexkilgour why do you hate `!default`? Out with it.

jpw: @davidpauljunior do you remember the `!default` reasoning?

packages/global-dropdown/scss/50-components/_global-dropdown.scss
sonniesedge: I would still assign this as a variable with a default value and allow it to be overridden later (but let's see what @alexkilgour has got to say about hating `!default`).

PR #64, opened 2018-07-23
Fix testing of monorepo packages

alexkilgour:
Attempts to fix an issue with running tests across all monorepo packages, where dependencies within individual packages are not installed before tests are run (they were being boostrapped using Lerna at the publication step). If a new dependency is introduced that is required in one of the tests then the build will fail as the dependency will not have been installed.

The solution to this is to run `lerna boostrap` _before_ running the tests (so before the build step). The [hoist command](https://github.com/lerna/lerna/tree/master/commands/bootstrap#--hoist-glob) is used to share deps across packages. But this introduces another problem, which is that validation will fail if dependencies are installed locally to a package as the `node_modules` folder is not a valid folder, and `package-lock.json` is not a valid file.

To solve these issues, this PR does the following:
- adds boostrap step to the Travis before the build step
- remove boostrapping from the publication script
- parse the projects `.gitingore` file and exclude matching paths from the validation step

Aside: As part of this ticket I experimented with replacing Lerna with [Yarn Workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/), which promised greater simplicity and speed. However, I found that I did not see the speed increases that I was expecting, and in fact using workspaces was considerably slower than using Lerna. In fact, Lerna experimented with switching to Yarn under the hood but ended up reverting back to NPM after the publication of NPM5. This issue gives an interesting summary of their reasoning: https://github.com/lerna/lerna/issues/1349


package.json
```diff
@@ -24,10 +24,11 @@
     "testfile": "jest --colors --verbose --silent",
     "lint": "npm run lint:js",
     "lint:js": "eslint **/*.js",
+    "bootstrap": "lerna bootstrap --hoist",
+    "build": "npm run lint && npm run test",
     "create": "sn-package-create",
     "validate": "sn-package-validate",
-    "publish": "sn-package-publish",
-    "build": "npm run lint && npm test && npm run validate -- -n"
```
alexkilgour: Had to remove the validation step until the `.gitignore` changes have been published otherwise this will keep failing. This will be added back in once a new version of `global-package-manager` has been published, and is in use

README.md
```diff
@@ -7,7 +7,17 @@
 
 The global toolkit contains front-end packages that can be used on all Springer Nature products. Packages can contain one or more of `css`, `javascript`, `templates`, `images`, `tests`. Each package should have its own folder in `./packages`, which result in it being published to NPM using the `@springernature` scope.
 
-### Writing a package
+## Installation
+
+The global toolkit functions as a [monorepo](https://medium.com/@maoberlehner/monorepos-in-the-wild-33c6eb246cb9), and uses [Lerna](https://lernajs.io/) to manage dependecies within it's packages.
```
josebolos: "dependecies within it's packages" -> "dependencies within its packages"?

PR #65, opened 2018-07-24
creates some utility classes

o-t-w:
Creates some utility classes. One file is simply entitled `utilities.scss` due to the fact that it contains miscelleaneous classes.


o-t-w:
Its a hard call. If a single class doesn't fit in anywhere, I don't feel like it warrants an entire folder all by itself. Naming things is hard :(

davidpauljunior:
I agree on the spacing comment.
I would suggest using a mixin that outputs the classnames and uses the values that are defined in a settings map (which can be redefined by a product or brand).
Padding could also be in there.

davidpauljunior:
Yeah you're right, naming stuff is very hard.

o-t-w:
As more classes get added over time the classes in `utilities.scss` may fit in somewhere else and maybe we can get rid of it.

o-t-w:
While I understand Jon's comment, defining it in a map feels hazardous to me. If you use the margin classes inside a global component, it could end up looking funky as the margins will be different from what you saw when developing the component. I think margin utility classes are really useful. However, _**if**_ they are brand dependent, they should be defined as utility classes **_inside the individual product repos_** and omited entirely inside the shared utility library imo. Having inconsistent utility classes will complicate things and confuse me.

davidpauljunior:
This is an interesting one.  Begins a discussion about whether components should be using utilities in them.  That creates a dependency.  Which, of course, may be ok.
I don't think a map would cause any problems, and a mixin would ensure the output classes are always consistent across projects.  But, it's entirely subjective.

o-t-w:
I think utilities for classes that can be added/removed/toggled from elements with javascript are useful inside components. I was spured to create this repo because I needed a `u-display-none` and `u-rotate` when building a component. Utility classes for margins are less necessary w/i a component though as we can do everything with BEM.

alexkilgour:
I feel like some stuff has been missed here. As per the [CSS guidelines](https://github.com/springernature/frontend-playbook/blob/master/css/how-we-write-css.md#60-utilities) ALL utilities that are more than a single rule should have an associated mixin.

Also, are we sure that those margin values are shared across all of our projects? otherwise they have no place in a global component.

And I think the name should reflect that these are global **css** utilites - so `global-css-utilites` rather than `global-utilites`.

Overall I think the version should have been kept at `0.0.0` whilst this is being worked on, as I don't think this is ready to be published to NPM.

o-t-w:
> ALL utilities that are more than a single rule should have an associated mixin.
What do you see as the benefit of this? I find that it makes the code less readable - its less easy to see what the CSS is actually doing at a glance.

o-t-w:
@alexkilgour I made an open space issue. https://github.com/springernature/frontend-playbook/issues/251

PR #66, opened 2018-07-24
Utility

o-t-w:
Name change for utility class package. I think it would be good to keep this small to start off with so that it can organically grow as people add utility classes that they find they actually need when working.


packages/global-css-utilities/scss/70-utilities/spacing.scss
```diff
@@ -0,0 +1,71 @@
+/* margins */
```
benjclark: does this comment need to be deleted? noted that there are no top line comments in the other files committed

packages/global-css-utilities/scss/70-utilities/utilities.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.u-transparent {
+	opacity: 0;
+}
+
+.u-rotate {
+	transform: rotate(180deg);
+}
+
+.z-index-4 { z-index: 4; }
```
benjclark: should this be `u-z-index` or `u-zindex` so we have consistency in naming and also so people can guess that classname without having to look it up?

o-t-w: Oh no. Nice catch Ben

PR #68, opened 2018-07-30
import normalize and very minimal base styles

o-t-w:
very minimal base styles


packages/global-css-base/40-base/normalize.scss
```diff
@@ -0,0 +1,341 @@
+/*! normalize.css v8.0.0 | MIT License | github.com/necolas/normalize.css */
+
```
alexkilgour: Is this just normalize v8 hard-coded in here? could we manage this as a dependency to make updating quicker/easier?

o-t-w: An NPM install of Normalize only has a .css file, not a Sass file, so its not possible to import it.

alexkilgour: ah, that's annoying!

packages/global-css-base/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@springernature/global-css-base",
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "normalize and base styles",
+  "keywords": [],
+  "author": "Ollie Williams"
```
benjclark: maybe we should make these `Springer Nature`? as people will be working collaboratively to improve these as time goes on

packages/global-css-base/scss/40-base/normalize.scss
```diff
@@ -0,0 +1,341 @@
...
+ */
+
+button,
+input,
+optgroup,
+select,
+textarea {
+  font-family: inherit; /* 1 */
+  font-size: 100%; /* 1 */
+  line-height: 1.15; /* 1 */
```
benjclark: this value has been declared more than once in this file. would both values benefit from being consistent with each other? if so, maybe we should variablise this at the top of the file so there is only one definition to manage and to ensure everything always stays consistent.

o-t-w: We could change it to `inherit`?

benjclark: ah yeah that's true

davidpauljunior:
Re normalize: https://www.npmjs.com/package/node-normalize-scss

packages/global-css-base/scss/40-base/normalize.scss
o-t-w: Its weird that the author of Normalize used `inherit` for font-family but not for line-height. Maybe there is a reason for that, but I can't figure it out 🧐

o-t-w:
@davidpauljunior As far as I am aware the normalize NPM package is not from the official maintainer of Normalize. There is no guarantee it will be kept up to date. Its just somebody who has copied and pasted Normalize into a `.scss` file - which is exactly what I am doing myself here.

packages/global-css-base/package.json
o-t-w: I think for this particular package that makes sense - but in general it is useful to know the person responsible for the bulk of the code so that you can ask them questions about it if need be.

packages/global-css-base/scss/40-base/normalize.scss
benjclark: well if there is reason to not keep those values consistent with each other then it's understandable. and in this case perhaps some people don't want to keep them consistent and want to allow the flexibility to have different values for each. /shrug

o-t-w: Its probably best to avoid altering Normalize as if we need to update it we will need to copy and paste it again and will have to keep track of any changes we've made.

davidpauljunior:
@o-t-w well it's not quite the same.  The node normalize is just a copy, but it's updated in sync with every new normalize release.  Obviously not a big deal, but I'm guessing that repo is more likely to be maintained than our copied version.

PR #69, opened 2018-07-30
@jonstevens added autocomplete component

jon-stevens:
Ello

I've added a basic autocomplete component.


packages/global-autocomplete/scss/main.scss
```diff
@@ -0,0 +1,49 @@
+.c-autocomplete {
+  margin: 10px auto 0;
+  width: 380px;
+  height: 50px;
+  font-size: 20px;
+  border: 1px solid #eee;
+  padding: 5px 10px 5px 0;
+  text-indent: 15px;
+  font-family: 'Montserrat', sans-serif;
```
o-t-w: Nice work Jon :) Things that are brand-specific like font-family should be configurable rather than hard-coded in.

.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "presets": [
```
o-t-w: I think build tool things will be defined on a project-wide basis rather than for each individual component.

packages/global-autocomplete/scss/main.scss
jon-stevens: hey! hey!
yes, I didn't mean to leave that in.

.babelrc
jon-stevens: I added this because I want Jest to understand es6 modules when running the tests

alexkilgour:
I would like to see a demo here, as my gut feeling is that this can be just a javascript component and we don't need to supply any css with this. Products can make their own styling decisions.

packages/global-autocomplete/__tests__/test.js
```diff
@@ -0,0 +1,139 @@
+'use strict';
```
alexkilgour: try and name the test file with the same name as the file being tested - so in this case `__tests__/index.js`

jon-stevens:
@alexkilgour The styling is very basic. Just enough to get the suggestions in a list, and add some hover/focus styles.

screen shot 2018-07-30 at 16 46 33

o-t-w:
@alexkilgour I reckon the ux people should take a look and decide if it is something they are happy to be visually consistent across brands, as they are keen to standardise. James A keeps telling me that "the age of boutique design is over" 😄

jpw:
Firstly, awesome stuff, thanks Jon!
Secondly, this is early days so some of the gubbins around components publishing may be rough and/or in a state of flux.

Also, why is this PR changing `packages/global-package-manager/...` ?

jon-stevens:
Ello @jpw . I added a babel config file so that Jest would import es6 modules in tests. That slightly broke one of the global-package-manager tests, so I put in a fix.
Also think it would be good to go over some of the other tests and change require style imports to es6, as this is part of our style guidelines. At some point

jpw:
Ello!

> I added a babel config file so that Jest would import es6 modules in tests. That slightly broke one of the global-package-manager tests, so I put in a fix.

Agree on all that, but should be a diff PR. Also @alexkilgour is doing surgery on that bit at the mo, so it may get pulled out of this repo entirely -- he will have the latest on that

.travis.yml
```diff
@@ -13,7 +13,7 @@ gemfile:
 matrix:
   include:
     # Run tests in Node.js 6.x
-    - node_js: '6'
+    - node_js: '7'
```
jpw: what was the issue here plz?

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,32 @@
...
+		endPoint: 'https://www.something.com/something/autocomplete?q=',
+		timeout: 2000,    // OPTIONAL: Provides a limit for the amount of time to wait for response from the server
+		minCars: 1,       // OPTIONAL: Min character limit. Defaults to 1
+		inputDelay: 300,  // OPTIONAL: Delay in between keyup and sending request to server. defaults to 300
+		headers: {
+			Accept: 'application/json; version=2'
+		}
+ };
+
+ const auto = autoComplete;
```
jpw: ?

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,32 @@
...
+ auto.init(args);
+ auto.enable();
+```
+
+#Notes
+
+This is a simple autocomplete component and it presumes that the request will return some basic JSON. You may need to alter the generateSuggestions method if you expect a more complex response.
```
jpw: Please provide sample JSON

packages/global-autocomplete/js/index.js
```diff
@@ -0,0 +1,226 @@
+const autoComplete = (() => {
+	const containerClassName = 'c-suggestion-container';
+	const suggestionClassName = 'c-suggestion-container__suggestion';
```
jpw: these classnames should be params i think?

packages/global-autocomplete/js/index.js
```diff
@@ -0,0 +1,226 @@
...
+		suggestionList.dataset.component = componentName;
+
+		if (resultsLength > 0) {
+			suggestionListKeyEvents();
+			selector.addEventListener('keyup', inputKeyEvents);
+		} else {
+			data.push('no results');
+			selector.removeEventListener('keyup', inputKeyEvents);
+		}
+		data.forEach((suggestion, index) => {
```
jpw: suggest users should be able to pass a callback here to render their own HTML, e.g. maybe they want links?

packages/global-autocomplete/js/index.js
```diff
@@ -0,0 +1,226 @@
...
+			let suggestionEl = document.createElement('div');
+			suggestionEl.innerText = suggestion;
```
jpw: would rather we stripped tags, or allowed a small subset of tags e.g. `b` `i` `strong` and the like

jpw:
> I did talk to Alex about this yesterday but yeah I will stick this in a different PR.
> tbf this pr wasn't actually ready for review - I wanted to see if the tests and stuff were passing on Travis 

Cool, thanks again for having a go 😃

.travis.yml
jon-stevens: Node 6 does not support some es2016 features, one of which I am using. I'll remove this as well and maybe put in a separate PR

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,57 @@
...
+    // Update UI with error state
+    // You can optionally call myAutoComplete.disable();
+}
+
+const args = {
+    selector: '.c-autocomplete',
+    onSelect: onSelect,
+    searchError: onError,
+    endPoint: 'autocomplete?q=',
+    timeout: 2000,          OPTIONAL: Set a timeout for the fetch request, onError will be called if fetch request timeouts
```
jpw: please specify default values for these optional parameters ;)

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,57 @@
...
+    minCars: 1,             OPTIONAL: Minimum characters to be typed before request is sent
+    inputDelay: 1100,       OPTIONAL: Delay between keypress and request being sent
+    headers: {
+      Accept: 'application/json; version=2'
+    },
+    containerClassName: 'c-suggestion-container',
+    suggestionClassName: 'c-suggestion-container__suggestion',
+    callBack: callBack
```
jpw: as you have 3 callbacks, maybe call this onSuccess? OnResults? or is that a bad idea as it implies it is an event..? 🤔

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,57 @@
...
+    suggestionClassName: 'c-suggestion-container__suggestion',
+    callBack: callBack		
+};
+
+const myAutoComplete = autoComplete(args);
+myAutoComplete.enable();
+```
+
+#HTML
+<input type="text" autocomplete="off" class="c-autocomplete" />
```
jpw: implicitly close void elements please

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+#Notes
+This component uses an es7 feature, Object.values which will require a polyfill or shim for certain browsers if you are not transpiling your code.  --> https://github.com/es-shims/Object.values
```
jpw: `ES7`. Also, am torn as to whether to mention this or not, given everyone should be transpiling... I guess leave it in, it's good to warn people.

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
+'use strict';
+var autoComplete = require('../js/index.js');
```
jpw: `const`

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+			auto.enable();
+			fetchSpy = jest.spyOn(global, 'fetch');
+		});
+
+		afterEach(() => {
+			document.body.innerHTML = '';
+			fetchSpy.mockReset();
+			fetchSpy.mockRestore();
+			fetchSpy.mockReset();
+			fetchSpy.mockRestore();
```
jpw: duplicated

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+		describe('Typing into the text input', function () {
+			test('Should send request to the specified endpoint', async function (done) {
+				input.value = 'bur';
+				input.dispatchEvent(new Event('keyup'));
+
+				setTimeout(() => {
+					expect(fetchSpy).toHaveBeenCalledWith(url + 'bur', {"content-type": "application/json", "headers": headers});
+					done();
+				}, 900);
```
jpw: timeout seems long?

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+			fetchSpy.mockRestore();
+			fetchSpy.mockReset();
+			fetchSpy.mockRestore();
+
+			mockResultsCallback.mockReset();
+			mockResultsCallback.mockRestore();
+		});
+
+		describe('Typing into the text input', function () {
+			test('Should send request to the specified endpoint', async function (done) {
```
jpw: `async done => {` ... can be replaced in a few places in this module ;)

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+				}, 900);
+
+			});
+
+			test('Should invoke callback function from config', async function (done) {
+				input.value = 'al';
+				input.dispatchEvent(new Event('keyup'));
+
+				setTimeout(() => {
+						expect(mockResultsCallback.mock.calls.length).toBe(1); /// TE
```
jpw: `/// TE` ?

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+				minCars: 1,
+				inputDelay: 100,
+				headers: headers
+		};
+		let fetchSpy;
+		let input;
+
+		beforeAll(() => {
+			global.fetch = () => {
+				return new Promise(resolve => {
```
jpw: Does it make sense to use https://jestjs.io/docs/en/mock-function-api.html#mockfnmockresolvedvaluevalue here?
(this is in a few places throughout this module)

packages/global-autocomplete/__tests__/index.js
```diff
@@ -0,0 +1,120 @@
...
+						expect(mockResultsCallback).toHaveBeenCalledWith(expectedResults);
+					done();
+				}, 150);
+			});
+		});
+
+		describe('A bad response', function () {
+			test('Should call the error function provided in the config', async function (done) {
+				global.fetch = () => {
+					return new Promise((resolve, reject) => {
```
jpw: has this code been linted? I would expect eslint to holler about `reject` being unused...

packages/global-autocomplete/js/index.js
```diff
@@ -0,0 +1,205 @@
...
+			event.preventDefault();
+			if (suggestions().length > 0) {
+				suggestions()[0].focus();
+				selector.value = suggestions()[0].innerText;
+			}
+		}
+	};
+
+	const addSuggestionEventListeners = () => {
+		if (container() !== null) {
```
jpw: return early rather than indenting please

packages/global-autocomplete/js/index.js
```diff
@@ -0,0 +1,205 @@
...
+	let componentName = args.componentName;
+	let onSelect = args.onSelect;
+	let inputDelay = args.inputDelay || 300;
+	let requestTimeout = args.timeout;
+	let headers = args.headers || {};
+	let searchError = args.searchError;
+	let callBack = args.callBack;
+
+	const keyCodes = {
+		down: 40,
```
jpw: JS can't do enums, so instead I do;
```
	const KEY_CODES = Object.freeze({
		DOWN: 40,
		...
```

packages/global-autocomplete/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@springernature/global-autocomplete",
+  "version": "0.0.0",
```
jpw: I think this needs to be at least 0.0.1 to publish... @alexkilgour ?

jpw:
Can we look at this again please? It's nice! (should gitignore `.idea/workspace.xml` though)

jpw:
Tests now failing due to changes in Babel, adding a `.babelrc` to this package and amending the root `babel.config.js` [to grok rc files in sub-packages](https://babeljs.io/docs/en/config-files#subpackage-babelrc-files) should get them working again.

jpw:
Whaddaya reckon @alexkilgour  ?

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,65 @@
...
+#Example usage
+
+Import the JS and the SCSS files into your project.
+
+#JS
+
+```
+import autoComplete from 'global-autocomplete';
+
+const showResults = results => {
```
sonniesedge: I know this is an example, but I suggest `onResults` for consistency.

packages/global-autocomplete/scss/main.scss
```diff
@@ -0,0 +1,47 @@
...
+	&::-webkit-input-placeholder {
+		color: #ccc;
+	}
+
+	&::-moz-placeholder {
+		color: #ccc;
+	}
+}
+
+.c-suggestion-container {
```
sonniesedge: The examples use `.c-results-container`.

sonniesedge:
@jon-stevens This looks very lovely. Has it been used in anger on a project yet?

jon-stevens:
@sonniesedge 

I extracted this from something that I made for Submissions (CA). We are also looking to use it elsewhere very soon! Would be good to get it merged

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,65 @@
+# Autocomplete component
+
+Basic autocomplete function that listens to key events on a text input and fetches suggestions from a specified endpoint.
+
+
+#Example usage
+
```
alexkilgour: Need a space after all of the `#` in the titles (4 fo them)

packages/global-autocomplete/package.json
alexkilgour: update version to `0.1.0`

packages/global-autocomplete/scss/main.scss
```diff
@@ -0,0 +1,35 @@
+.c-autocomplete {
```
alexkilgour: the path/name for this should be `packages/global-autocomplete/scss/50-components/autocomplete.scss`

packages/global-autocomplete/README.md
```diff
@@ -0,0 +1,65 @@
...
+Import the JS and the SCSS files into your project.
+
+#JS
+
+```
+import autoComplete from 'global-autocomplete';
+
+const onResults = data => {
```
joeyciechanowicz: Expecting the consumer to have an endpoint that returns data in the expected format will add additional work for teams looking to add/replace an autocomplete with an existing backend.
Converting an array of items to DOM elements (with required attributes) and attaching them at the correct place, seems like a job for the module rather than the consumer.

Perhaps a more flexible solution is for the consumer to provide a `onInput` (or better named) function which is called by the module, and it will return a `Promise` which resolves to an array of items. This would allow it to be slotted in to existing code a bit easier (and provide transformations on the backend response).

packages/global-autocomplete/package.json
```diff
@@ -0,0 +1,11 @@
...
+  "license": "MIT",
+  "description": "a basic autocomplete tool",
+  "keywords": [],
+  "author": "Jon Stevens",
+  "devDependencies": {
+    "object.values": "^1.0.4"
```
joeyciechanowicz: `Object.values` has been supported since Node 7.0, and this project is on `Dubnium (v10.15.3)`

jon-stevens: Actually this is not needed anymore anyway - will remove

packages/global-autocomplete/README.md
jon-stevens: > Converting an array of items to DOM elements (with required attributes) and attaching them at the correct place, seems like a job for the module rather than the consumer.

This was initially my thought, but I changed it (after a previous comment on this PR) so that the DOM manipulation with the results is left to the individual user, as it quite possibly will vary. All this component expects of the response is that it is JSON (which could perhaps also be made configuarable)

jon-stevens: > Perhaps a more flexible solution is for the consumer to provide an onInput (or better named) function which is called by the module, and it will return a Promise which resolves to an array of items. This would allow it to be slotted in to existing code a bit easier (and provide transformations on the backend response).

👍 
Perhaps the consumer could pass in an optional onInput function; if it is passed in, then the component calls it, if it is not, then it uses the existing fetch method as is?
- I wanted this to be something simple that you can just feed a URL and basically be up and running.

jpw:
Happy Birthday to this PR 🍰 🎁

PR #70, opened 2018-07-31
add visually-hidden utility class

o-t-w:


packages/global-css-utilities/scss/60-utilities/hiding.scss
```diff
@@ -5,3 +5,10 @@
 .u-transparent {
 	opacity: 0;
 }
+
+.u-visually-hidden {
```
davidpauljunior: Should this `@include visually-hidden`?

o-t-w: IMO no. This way I can see exactly what is going on w/o needing to open the mixins directory and finding the right file.

davidpauljunior: But any code changes need to be done twice.

benjclark: It's important that code is maintained in one place as duplication has some obvious downsides like significantly increasing the risk that people forget to update both sets of code or adding extra time cost to the management of your code. I understand how annoying it is to follow a trail to find the code you want as well. Perhaps there is a best of both approach to this where we maintain the code in one place but try to find tooling (editor extensions for example) that allows a developer to at least see any `@include` code in the file that's calling it.

o-t-w: I just did some googling and couldn't find anything. Do you know of any such tool. It would be very useful :)

o-t-w: I've used the mixin

packages/global-css-utilities/scss/30-mixins/hiding.scss
```diff
@@ -0,0 +1,6 @@
+@mixin visually-hidden {
+	width: 1px;
+	height: 1px;
+	position: absolute !important;
```
davidpauljunior: Is the `!important` necessary?  I mean, why is it needed on the position, but not the width or height.

o-t-w: Oops! I copied it over from shunter-mosaic and didn't notice that

PR #71, opened 2018-08-03
@jonstevens added babel config

jon-stevens:
Added a .babelrc so that the test files can import es6 modules.
This broke one of the existing tests (once babel has done it's thing _this_  is now undefined within the tests - it seems you have to use _jest_ instead).


benjclark:
I swear you committed these changes in another PR recently as well?

jon-stevens:
I had committed them as part of another PR because I am berked. Pulled out into another PR

PR #77, opened 2018-08-14
Use .nvmrc

jpw:
This PR adds an `.nvmrc` and guidelines on how to use it.

This is because I was hitting problems due to differences in `npm i` results, due to differences in what version of `node` (and therefore `npm`) I was using locally. This is really opaque to developers.

By using an `.nvmrc`:
1. it should make these problems go away
1. we get reproducible builds
1. devs working in the repo will not forever be committing changes to the `package-lock.json` after they `npm i` with varying results, according to their local version of `node` and `npm`.
1. we get consistency in non-browser JS in the repo, such as some test code

As and when new minor versions of `node` are released, it will be up to someone to test with the new version, update the `.nvmrc`, run an `npm ci` then commit the changes to `.nvmrc` and `package-lock.json`. A small price to pay, in the grand scheme of things I think.


README.md
```diff
@@ -11,12 +31,22 @@ The global toolkit contains front-end packages that can be used on all Springer
 
 The global toolkit functions as a [monorepo](https://medium.com/@maoberlehner/monorepos-in-the-wild-33c6eb246cb9), and uses [Lerna](https://lernajs.io/) to manage dependencies within it's packages.
 
-To get started, install dependencies:
+To ensure that installs are consistent, we all use the same version of `node` & `npm` when working in the repo and use `npm ci` to install dependencies.
+
+### Using the correct `node` & `npm` versions
+
+Ensure you have [`nvm` installed](https://github.com/creationix/nvm/blob/master/README.md), then in the root of the repo run `nvm use`. This will read the version of `node` (and therefore `npm`) to use from the `.nvmrc` and ensure you are using the correct versions.
```
joeyciechanowicz: Do we want to note that you need to run `nvm install` first, if you haven't used that version of node before.

joeyciechanowicz: Though it does inform you that you need to do so after running `nvm use` anyway. So probably making this sentence un-necessarily verbose.

jpw: Good thoughts! Well, that para tries to accomplish 3 things:
1. you need nvm installed
1. you need to run it in the root
1. an explanation of what the .nvmrc is if you don't know, and how it basically works

Which bit do you think might be verbose? 😄

joeyciechanowicz: My proposed change would make it more verbose! I was back tracking on my first comment 😉 

Especially as running `nvm use` without first running `nvm install` (where necessary) will inform you to run `nvm install` anyway.

Scratch this!

jpw:
> What was the problem exactly.

I was hitting problems where i would `npm i` between versions 6, 8 & 10 and got different results (naturally) which was causing problems with running tests. Trying to use 6 was user error, somehow, which I couldn't figure out then gave up, as we should be using `.nvmrc` anyway I think. Also because it fixes other errors described in the orig PR message -- trying to debug this was a PITA with versions of underlying deps (and therefore the `package-lock.json`) changing all the time.

> Why 8.11.3 and not any other version?

Well 10 felt a bit new & shiny, 8.11.3 is the latest 8 LTS version

> How are we expecting to keep this version up to date?

Developer enthusiasm and/or need, same as everything else ;)

jpw:
FML `npm ci` is in npm 5.7.0, but `node v8.11.3 (npm v5.6.0)`
So would need to hike it all to node 10 anyway 😝shall we just try node 10 and see how we go?

jpw:
> Node.js 10.0.0 is the seventh major Node.js release since the launch of the Node.js Foundation. In October of 2018, it will become the next Active Long Term Support branch.

https://nodejs.org/en/blog/release/v10.0.0/

.nvmrc
```diff
@@ -0,0 +1 @@
+v8.11.3
```
alexkilgour: Does this enforce this _exact_ version of node? or is it possible to specify a range?

README.md
```diff
@@ -1,5 +1,25 @@
...
+        - [SASS](#sass)
+        - [Javascript](#javascript)
+    - [Testing](#testing)
+    - [Linting](#linting)
+    - [Continuous integration](#continuous-integration)
+        - [Publishing](#publishing)
+    - [License](#license)
+
+<!-- /TOC -->
+
```
alexkilgour: Think the TOC should be below the short description, so just about the `installation` heading, so you don't have to scroll down to see what the repo is for and see the badges

.nvmrc
alexkilgour: or this desired that we fix to a very specific version?

jpw: we could do e.g. `v8.11`

alexkilgour:
You can update `npm` independently of `node`, it is done in the `travis.yml` file so that `npm ci` can be used on travis. Not sure if you can do this using NVM though

alexkilgour:
Also, if we bump everything to 10, need to make sure it is done for travis as well

jpw:
> You can update npm independently of node, it is done in the travis.yml file so that npm ci can be used on travis. Not sure if you can do this using NVM though

It's doable locally but not through nvm, it's a minor faff but I assume that versions of node are tied to versions of npm for a reason _waves hands around_

jpw:
> Also, if we bump everything to 10, need to make sure it is done for travis as well

Is it a problem doing this for travis?

alexkilgour:
Not sure. I would imagine not, but worth checking...
Quick google throws up some people with issues, but should be able to get around it
e.g. https://github.com/nodejs/node-gyp/issues/1420

alexkilgour:
also, we can stick a lovely little badge at the top to specify versions :-)



`https://img.shields.io/badge/node-8.11.x-brightgreen.svg`



`https://img.shields.io/badge/nvm-%3E=5.7.0-brightgreen.svg`

jpw:
Eesh, well as 10 is not LTS yet, suggest I amend this PR to:
1. pin to a minor not patch version
1. rm references to `npm ci`
1. move toc

... and then revisit node 10 in October..?

jpw:
Alex and I just had a chat about this.
We have elected to go back to using `npm i` for now and seeing how we go, hopefully if devs use `nvm use` we should minimise changes to the lock file. We can revisit when node 10 is LTS and also better supported in Travis, then maybe switch to `npm ci`.

Also removing the node bits in the travis file as [travis parses `.nvmrc` files](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc)

josebolos:
Hello! Sorry again, I have another question, as I'm not familiar with how the components are supposed to work. What is the reason behind only testing this in only one version of node? Are we expecting everyone to use exactly the same version in production? How does this work?

alexkilgour:
The components themselves are just client-side code to be used in projects (HTML, CSS etc) so there is no `node` requirement to use them, but we require node in order to run the command line utilities that support the repo - testing, linting, static analysis - and it makes sense for the reasons above for everyone to be running the same version to get reproducible results

josebolos:
Oooh! I understand now!

chris-pratt

(That said, I see no reason _not_ to use Node 10)

PR #83, opened 2018-08-23
make dropdown slightly wider. Stop page scrolling when navigating wit…

o-t-w:
…h keyboard.


packages/global-dropdown/HISTORY.md
```diff
@@ -4,4 +4,7 @@
     * First version
 
 ## 0.1.1 (2018-07-20)
-    * Publish to NPM
\ No newline at end of file
+    * Publish to NPM
+
+## 0.2.0 (2018-07-23)
```
josebolos: I thought we write history from the most recent to the oldest one? e.g. https://github.com/springernature/frontend-global-toolkit/blob/master/packages/global-grid/HISTORY.md

Other than that, looks fine to me! 👍

o-t-w:
@josebolos oh yeah! Fixed.

PR #84, opened 2018-08-30
removes some things that should probably be at brand level. Adds box-…

o-t-w:
…sizing border-box


packages/global-css-base/scss/40-base/base.scss
```diff
@@ -1,12 +1,24 @@
 @import 'normalize.scss';
 
+html {
+	box-sizing: border-box;
+	font-size: 62.5%;
```
sonniesedge: Why this? It seems unnecessarily opinionated.

o-t-w: @sonniesedge This is used on Nature.com and BMC. It is a pretty common practice as it makes it easy to understand what an em unit will equal in pixels. 1rem is 10px, 1.1rem is 11px, 1.2rem is 12px, and so on.

o-t-w: @davidpauljunior I purposefully left out resetting the body font-size to 16px. On nature.com it is 17px. It will need to exist at brand level as it is not consistent across sites.

o-t-w:
I just noticed that BMC links have an underline by default, unlike most of the other websites, so I have removed some CSS. 
Also `text-decoration-skip: ink` (or `text-decoration-skip-ink: auto`) is the default in all browsers that support it, so it is pretty pointless including that code here.

davidpauljunior:
Hmmm, but a `$base-font-size` variable would fix that? Without resetting to a sensible number, the `62.5%` is quite dangerous.

Nature `$base-font-size: 1.7rem;`, BMC `base-font-size: 1.6rem`.

o-t-w:
@sonniesedge How would this commit prevent you from doing that?

davidpauljunior:
If projects are using rems based upon a root size of 16px, rather than 62.5%, then this is going to completely screw those people.  Even setting it back to 1.6rem (or whatever) on the `body` won't help, as all references to rem units would suddenly be based on a different value than previously.

Given that, it seems unavoidable to remove that bit of code from this file.

o-t-w:
@sonniesedge the font sizing has been removed.

o-t-w:
I don't see the point in duplicating the code - the box sizing I removed from Normalize will be achieved by our global selector instead. Even if somebody were to copy-paste Normalise again to get an update & didn't make the same changes, nothing would break, so I don't see the problem.

davidpauljunior:
If nothing would break by making those changes I'd say that leaving the external dependency in it's original state is more sensible.

packages/global-css-base/scss/40-base/base.scss
```diff
@@ -1,12 +1,17 @@
 @import 'normalize.scss';
```
sonniesedge: I'd also ask why this is here. Doesn't sub-importing in this way go against the layered method that we use elsewhere?

On Lib Portal normalize is used, but it's pulled in as part of the advanced stylesheet.

jpw:
Is this PR still up for discussion?

davidpauljunior:
@jpw this PR is so old I'd be inclined to close it and not merge it?

PR #88, opened 2018-10-01
Update deps and fix readme issues

alexkilgour:
- update the dependencies (and package-lock.json)
- fix grammatical error in readme
- remove example gif (it appears in the package readme already)


alexkilgour:
@o-t-w Sorry, I meant I have removed it completely from this repo - so the image and from the readme, because the example appears in https://github.com/springernature/frontend-package-manager

PR #92, opened 2018-10-16
Sasslint

o-t-w:


.sass-lint.yml
```diff
@@ -0,0 +1,92 @@
+options:
```
alexkilgour: Where did the rules come from?

alexkilgour: ah, think it's from the playbook right?

.sass-lint.yml
```diff
@@ -0,0 +1,92 @@
...
+    space-between-parens:
+        - 2
+        -
+            include: false
+    space-around-operator: 2
+    trailing-semicolon: 2
+    indentation:
+        - 2
+        -
+          size: 4
```
alexkilgour: this appears as `size: 'tab'` in the playbook, is it intentionally different here?

alexkilgour: if you change this to tab then i think this fixes some of your 'mixed tabs and spaces' issues - the rest are genuine issues!

alexkilgour:
I would also like to have a shared config that we can use across the projects, instead of relying on replicating what we have in the playbook each time - this would mean creating something similar to `eslint-config-springernature`, so `sass-lint-config-springernature`

take a look at this for some hints: https://github.com/sasstools/sass-lint/issues/718

PR #99, opened 2019-01-15
dubnium updates

jpw:


benjclark:
Oh, actually, just thought of something. Doesn't the node version need to be updated in the engines -> node field in package.json ?

jpw:
hmm yeah thanks, I always ignore that, as does most software... will see if i can remove it

alexkilgour:
As far as I am aware the `engines` field is advisory, so can be removed

jpw:
also had to rm some travis config to run the tests on all branches... seems to be a historic restriction to do with semver releases candidate branches (or similar). We want to run the tests on all branches so have removed that bit of the travis config.

PR #102, opened 2019-01-30
form input component

sky-jack:
This is a speculative initial commit for a global form component. This is close to what I have been using in identity and thought it might work as a useful more general component, but was hoping to get some initial feedback on the approach from interested parties. In case I'm missing something important.

image


huckocs:
This won't be helpful review wise, but wanted to say this is a great start! Look forward to seeing it get approved and appearing in Elements!

packages/global-forms/view/form.html
```diff
@@ -0,0 +1,22 @@
+<div class="c-input-group">
+	<label for="test-input-0" class="c-input-group__label">input</label>
+	<input id="demo-input-0" name="test-input-0" type="input" class="c-input-group__control" autocomplete="off" value="" required="true" aria-required="true">
```
jpw: `<input ... type="input"` ??

Also if `required` (no need for `="true"` IIRC) I would expect some kind of indicator saying it's a required field, or whatever UX would suggest there.

Also we have CSS for textarea's and select's but no demos of those input types. Some radios might be nice too?

packages/global-forms/README.md
```diff
@@ -0,0 +1 @@
+This module gives you some generic styles for form inputs
```
jpw: Maybe specify which input types are covered at present?

packages/global-forms/scss/50-components/_global-forms.scss
```diff
@@ -0,0 +1,71 @@
...
+.c-input-row {
+	display: flex;
+	flex-wrap: wrap;
+	flex-direction: row;
+	margin-right: -$global-form-grid-gutter-width / 2;
+	margin-left: -$global-form-grid-gutter-width / 2;
+
+}
+
+.c-input-row .c-input-group {
```
sonniesedge: I think this is unnecessary nesting. It raises the specificity of the selector.

sonniesedge:
THANK YOU FOR DOING THIS JACK! ❤️ 

- My initial thought is about the fact this is in the global level: as I understand things, the forms would need to be implemented for each brand. 
- Screenshots in the PR would be extremely helpful for visual folks to check this out.
- As I read the code, this would result in labels horizontally next to inputs. Is this the best usability and accessibility practice?

packages/global-forms/scss/10-settings/_global-forms.scss
```diff
@@ -0,0 +1,17 @@
...
+ * @springernature/global-forms
+ * Default skin settings
+ * Replace these in your application
+ *
+ */
+$global-form--padding: 12px;
+$global-form--border-color: #999;
+$global-form--border-radius: 0;
+$global-form--background-color: #fff;
+$global-form-background-color-disabled: #f3f2ef;
```
davidpauljunior: Is there a reason for some of these to be `--` and others `-`?
I personally prefer just the `-`.

packages/global-forms/scss/50-components/_global-forms.scss
```diff
@@ -0,0 +1,71 @@
...
+.c-input-group textarea[disabled] {
+	background: $global-form-background-color-disabled;
+	border-color: $global-form-border-color-disabled;
+	color: $global-form--color-disabled;
+	cursor: not-allowed;
+}
+
+.c-input-group input[disabled]:hover,
+.c-input-group textarea[disabled]:hover {
+	background: $global-form-background-color-disabled;
```
davidpauljunior: These rules seem to repeat those in the non `:hover` state, could they be grouped with them?

packages/global-forms/scss/50-components/_global-forms.scss
```diff
@@ -0,0 +1,71 @@
+
+.c-input-group {
+	display: block;
+	margin-bottom: 1rem;
+}
+
+.c-input-row {
```
davidpauljunior: This feels to me like it shouldn't be part of the forms and that layout should be controlled in a separate component, for example a `c-flex` or something.

packages/global-forms/scss/50-components/_global-forms.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.c-input-group__required {
+	color: $global-form--color-required;
+}
+
+.c-input-group__control {
+	@include global-form-input
+
+	&--check {
```
davidpauljunior: I wonder if there's a lot more to override here from the `global-form-input` mixin. E.g. border, background-color, padding, border-radius.
Perhaps, radios and checkboxes need to be considered different things to your `.c-input-group__control`?

packages/global-forms/scss/50-components/_global-forms.scss
huckocs: Wouldn't flex be more a utility class?  -  u-flex
Is there such a thing as a layout class?   -  l-flex

Check me contributing.

davidpauljunior: Haha that's a whole other debate. We made it a component because we have child class names that change the `flex` values, but it does feel a bit utility-ish

sky-jack:
Thanks for all the feedback people. I've just pushed a change to improve the demo markup and tweak the radio buttons. Further changes on the way.

packages/global-forms/scss/10-settings/_global-forms.scss
sky-jack: No' I'll change that. The existing convention seems to be for  '--' so I'm going to stick to that I'm afraid.

packages/global-forms/scss/50-components/_global-forms.scss
sky-jack: yes I guess so, I'm loath to have another class the for row item though but could do that.  if we hand this off to a flex utility then we can definitely remove, any other ideas how to achieve this?

packages/global-forms/scss/50-components/_global-forms.scss
sky-jack: Yes quite possibly

packages/global-forms/scss/50-components/_global-forms.scss
sky-jack: I suppose they could. I guess my logic for including this was that having multi-row forms is a fairly common use case. Reliance on a flex utility wouldn't create any dependencies and likely to be widely used, so yeah guess that could be the way to go.

packages/global-forms/scss/50-components/_global-forms.scss
sky-jack: You mean on the checkbox element? 

My thoughts are they should remain the same. conceptually the control element covers all inputs.   Would you see it as something like:      `.c-input-group {}  .c_input-group__input{} .c_input-group__checkbox{}  .c-input-group__textarea {}` ?

sky-jack:
@sonniesedge Cheers :) .  re suggestions: The settings file in this package would be replaced at by brand-specific variables in each project. So the settings file in the package is more for demonstration purposes. I've added a screenshot that should help illustrate what is going on.

jpw:
I am loving the 🐟 direction this PR has taken.

I have dismissed my review as Jack has kindly addressed my quibbles, but others should look at the rest I expect.

packages/global-forms/scss/50-components/_global-forms.scss
davidpauljunior: Actually I just realised on BMC we use a modifier on the parent and break the single level specificity of BEm.  I think that's because a radio or checkbox _must_ be `input[type="checkbox"` or `input[type="radio"`.

```
.c-form-field--option {
    input[type="checkbox"],
    input[type="radio"] {
        margin-right: 0.5em;
    }

    .c-form-field__label {
        display: flex;
        align-items: baseline;
    }
}
```

I have absolutely no idea if that is the right way to approach it though 😭.
Forms are so complicated 😞

davidpauljunior:
Kudos @sky-jack for tackling forms, probably the hardest component to do.
I suspect that debate on the _correct_ way to do this could (and absolutely shouldn't) go on indefinitely.  There's always room for tinkering after the component is merged of course.

Probably worth getting @alexkilgour to look through and provide some feedback?

jpw:
Ooh yes, probably worth asking @hollsk to have a look too.

packages/global-forms/scss/50-components/_global-forms.scss
sky-jack: I've refactored this so that selector is less specific.   It means another class for `c-input-row__item` but I think that's good. I guess we could drop the row concept altogether but I'm inclined to keep it in to reduce dependencies on grid systems or suchlike.

packages/global-forms/view/form.html
```diff
@@ -0,0 +1,51 @@
+<form>
+	<fieldset>
+		<div class="c-input-group">
+			<label for="test-input-0" class="c-input-group__label">Email</label>
```
hollsk: Agree with @jpw that where a field is required, there should be something visible in the label text to indicate that

packages/global-forms/view/form.html
```diff
@@ -0,0 +1,51 @@
...
+			<input id="demo-input-0" name="test-input-0" type="text" class="c-input-group__control" autocomplete="off" value="" required aria-required="true">
```
hollsk: I think that we need to be careful with putting `autocomplete="off"` in here because it can prevent form fillers from working correctly, which is a pain for users with, say, memory issues or for people who struggle to type quickly. 

I know that this is going to be a generic component, so I don't want to get all nitpicky on the specific form fields being used here, but [anything in this list](https://www.w3.org/TR/WCAG21/#input-purposes) (that also meets the use case of being about the user) should almost certainly not have autocomplete disabled. I'd worry about people applying the property indiscriminately in their own components because they see it in the generic one, without realising that it can cause problems for people.

packages/global-forms/view/form.html
hollsk: The labels are trying to associate with the wrong attribute in the inputs (should be `demo-input-n`, not `test-input-n`)

packages/global-forms/view/form.html
```diff
@@ -0,0 +1,51 @@
...
+			<label for="test-input-4" class="c-input-group__label">Describe your fish</label>
+			<textarea rows="4" id="demo-input-4" name="test-input-4" class="c-input-group__control c-input-group__control--textarea" autocomplete="off" value=""></textarea>
+		</div>
+
+		<div class="c-input-group">
+			<label for="test-input-5" class="c-input-group__label c-input-group__label--check">Bottom feeder?</label>
+			<input id="demo-input-5" name="test-input-5" type="checkbox" class=" c-input-group__control c-input-group__control--check" autocomplete="off" value="" required aria-required="true">
+		</div>
+
+		<div class="c-input-group">
```
hollsk: Axe will complain about this pattern forever because they prefer that radio groups are identified with some kind of text label, above and beyond the labels for each individual input - by itself a radio group doesn't necessarily give the user much information on what the control is for. 

A radio group is a good place to start a new fieldset (with legend!). [Deque have more suggestions here](https://dequeuniversity.com/rules/axe/3.1/radiogroup?application=AxeChrome) but a fieldset/legend for each radio group is the simplest.

hollsk:
I love it. It's glorious. 

I'm almost scared to say it because of the amount of scope creep involved but... what about error messages? 😱 😱 😱 (could probably be done in a different component tho 😅)

sky-jack:
Cheers Hollie, 👍 I'll get on to these.

sky-jack:
Do you mean field-specific errors? yes a different kettle of fish, for client side errors I'm inclined to rely on html5 basic constraints. Guess the display of server-side errors is more of a UX question that could concievably change between brands (but perhaps should not)

packages/global-forms/view/form.html
sky-jack: Yes I see what you are saying.  On ID We've removed these in circumstances when every field was required as currently all our fields are required. IE on a login form or minimal registration form it seemed clearer to indicate in the text above that all fields are required, in other cases a graphical indicator would definitely be useful and from an accessibility perspective as well perhaps..

packages/global-forms/view/form.html
sky-jack: Are there security implications to consider in the use of autocomplete? I have vague memories about this but I'll have to look it up.

jpw: Autocompletion is complicated by whether the field is for PII or sensitive information, and what browser vendors think. The [MDN article about turning off autocompletion](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion) is nice, which I reference in the bit about autocompletion in the [frontend playbook secure markup](https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md#a-note-on-autofill) page. Which isn't very well-written, now I look at it again.

tldr; I'd probably remove the attribute altogether as it will be quite use-case specific, plus what Hollie said which was very interesting.

packages/global-forms/view/form.html
sky-jack: Thanks that is a useful link.

packages/global-forms/scss/50-components/_global-forms.scss
alexkilgour: Sorry, I know I am rather late to this. IMO you could do this either with utility classes or as Jack has done it here. I'm inclined to think it makes sense here because as Jack says it is key to how we want forms to look. In the spirit of keeping things "isolated and modular" this makes sense, rather than having to specify to people what utility classes to use in order to achieve this - simplicity is better than extra complexity for the sake of avoiding some repetition.

alexkilgour: Also with regards to flex feeling more like a utility but breaking some of the "rules". The rules are not 100% rigid, in some cases things will not fit into a nice little box and we shouldn't tie ourselves in knots trying to make everything conform 100%. Treat them as guidelines more than rules, and try and stick within them, but in some cases we can bend the rules a bit.

packages/global-forms/HISTORY.md
```diff
@@ -0,0 +1,4 @@
+# History
+
+## 0.1.0 (2019-01-30)
```
alexkilgour: Should start with v `0.0.1`

packages/global-forms/package.json
```diff
@@ -0,0 +1,11 @@
+{
+	"name": "@springernature/global-forms",
+	"version": "0.0.0",
```
alexkilgour: Version to match `history.md`

packages/global-forms/scss/40-base/_global-forms.scss
```diff
@@ -0,0 +1,29 @@
...
+
+legend {
+	margin-bottom: 0.5rem;
+}
+
+fieldset {
+	border: 0;
+	border-bottom: 1px solid #{$global-form--border-color};
+	margin-bottom: 1rem;
+}
```
alexkilgour: Alex/Jack: Base styles should go in the global/brand context, not in the component. Thinking of removing the `legend` and `fieldset` styles and leaving it to specific implementations. `textarea` seems useful and non-controversial so we can move that to context (need to check this). `label` style here can be part of the component class, and `input/select/textarea` styles look like they could be shared in context.

alexkilgour: Alex/Jack: todo - remove styles here for `fieldset` and `legend`. Move `label` styles to the component class. Put a note at the top of this file to move this into the global context once it exists.

morgaan:
Hi! First off I would like to +1 on @sonniesedge comment "THANK YOU FOR DOING THIS JACK! ❤️".

I was actually initiating work on that topic when I got aware of your PR.

On a general note, I would keep this global form module simple stupid (KISS FTW) and avoid the layouting like grid... to sneak into it.
Integrators are then free to use whatever utility classes that suits their needs. Although I would warn not to go in the direction of aligning form group horizontally.
You may have to add more code to address responsiveness of the form. In terms of internationalisation you would need to swap (more code maybe) the fields for right-to-left readers, where both left-to-right and left-to-right readers won't be affected by fields stacking on each other by default (universal design). The last point also applies to the label+input pair, that should be also kept vertically aligned.

I would add the notion of hint which of course is optional depending of the complexity of the input. This hint should be nested inside of the label so that it would be spoken by screen reader when focusing on the field.

I would also add the notion of error to also be nested inside of the label. Here is an example of a form failing validation taken from the great book "Form design patterns" from Adam Silver http://nostyle.herokuapp.com/examples/validation

In regards to required fields, there is this affordance of using * to denote required field. This might not be understand by every one, are someone may also miss the legend for it. More resilient option is to, depending on your use case, to suffix the label with `(required)` or `(optional)`. I believe it is more likely that most of the fields are required rather than optional (if not you may reconsider why and how you can shorten your form for ux sake), therefor the `(optional)` option sounds the most relevant. Anyway as long as the `required` html attribute is positioned on the field this should be spoken/shown to screen readers users.

I'm pretty motivated to bring this project (setting the guidelines about building forms) forward. This can and should constitute one of the early contribution of the design system I believe. I do not know your currenty availibility and/or interest on this topic. But would could totally work on it together ☀️

jpw:
Thanks for having a look @morgaan ! Jack is currently 1 day/week, so his time is limited, how would you like to take this forward Jack?

My 2p on some of the issues you discuss:

- **layout** I agree in principle, but if UX have guidelines on how to layout forms, I think it would be a nice-to-have as a default -- I'd appreciate not having to think about how to lay out a form as a developer. Users of the component are free to take the HTML or not, so as long as the CSS in this component is not going to cause problems, I would like a default layout.
- **i18n** you are correct, but rtl languages are not something we have to consider here AFAIK. I don't know if this could be built in a text-direction agnostic way, if it could, that would be nice.
- **hints** nice idea!
- **errors** also a nice idea, IIRC it's good practice to have errors near the erroneous field, rather than in a box at the top or bottom of the form. This might be very implementation-specific and hard to design though 🤔 agree with you about using browser validation too.
- **required or not being the default, and how to indicate that**... UX will have theories on that I'm sure, @huckocs ?

sky-jack:
Hey @morgaan Thank you for all your feedback on this. Lots to think about.

Alex and I were thinking that it might be a good idea to get what we have merged into master. Then people can begin to make use of the package. Once we are at that stage, we can open issues for each of your comments to discuss and implement suggested changes one by one. What do you think?

jpw:
As per Jack's previous comment, let's merge and go from there.

PR #103, opened 2019-02-04
Make media queries relative and not require includePaths

alexkilgour:
The previous change to this component broke people's builds as it required `includePaths` to set the location of installed dependencies for sass `@import` statements. This feels like an unnecessary step for a component dependency. This PR changes `@imports`'s to be relative (there is no reason you cannot _choose_ to use `includePaths` in your application, this change is compatible with that choice, but we don't enforce it).

Making this change requires that we know where dependencies will be installed (NPM will always _try_ and hike the dependency to the top level). To achieve this we are using `peerDependencies`, which ensures component and dependency always exist at the same level.


davidpauljunior:
Mighty fine work

PR #105, opened 2019-03-06
global-box: update naming and add template

alexkilgour:
- add handlebars templates
- update naming to meet current conventions


packages/global-box/view/box.hbs
```diff
@@ -0,0 +1 @@
+<div class="c-box">{{content}}</div>
```
joeyciechanowicz: Does content want to be escaped?

Alternatively, do we also want to use partials that allow content for this sort of thing?
```
// Template
<div class="c-box">
    {{> @partial-block }}
</div>

// Usage
{#> box }}
  My Content
{{/box}}


``

Comment (alexkilgour) [reply packages/global-box/view/box.hbs]: yeah, could do that. let me take a look.

Comment (joeyciechanowicz) [reply packages/global-box/view/box.hbs]: For simple components (like this one) one could support both methods as well.

```
{{#if content}}
<div class="c-box">
   {{content}}
</div>
{{#else}}
<div class="c-box">
    {{> @partial-block }}
</div>
{{/if}}
```

alexkilgour: could do that, but this is just an example implementation, so I think an example of how `@partial-block` works should suffice, which I have now added 🎉

PR #109, opened 2019-04-26
Add global article

saramorueta:
@nickcall @abdallaomar and I have moved the package`nature-article` from https://github.com/springernature/frontend-nature-toolkit to become our new global-article as a first step.

We have tried to keep the changes log for each file so the process has been a bit complicated. Even though, it should be safe to merge it onto master. (I think 🙏 )

We have decided not to move the package `nature-context` for now, until seeing the whole picture around it and decide how/what to do.


packages/global-article/HISTORY.md
```diff
@@ -0,0 +1,61 @@
+# History
+
+## 0.0.18 (2019-04-26)
```
alexkilgour: you need to wipe the history, as this is published as a completely new component

packages/global-article/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@springernature/global-article",
+  "version": "0.0.18",
```
alexkilgour: make this `0.0.0`, otherwise it will be published and we don't want that yet

packages/global-article/package.json
```diff
@@ -0,0 +1,9 @@
...
+  "description": "Frontend package for articles",
+  "license": "MIT",
+  "peerDependencies": {
+    "@springernature/nature-context": "^0.0.8"
+  }
```
alexkilgour: remove `peerDependencies` completely for now until we have `global-context`

alexkilgour:
You don't need to keep the commit history of the `nature-article`, that will still exist in the nature repo, treat this as a completely new component

saramorueta:
wow! I thought we would forget about `nature-article`. Now I'm confused 😕

packages/global-article/package.json
saramorueta: Sorry, We don't understand this. We were planning to start using it on nature and BMC.

saramorueta:
This is not necessary anymore. this other PR will add the new global article component.
https://github.com/springernature/frontend-global-toolkit/pull/110/

PR #136, opened 2019-05-20
hikes lerna

jpw:


davidpauljunior:
I take it the `~` is intentional?

jpw:
it was not, thanks David :)

PR #144, opened 2019-05-30
Adapt c-article-subject-list to support keywords for bmc and some nat…

moddular:
…ure articles


packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -1036,30 +1036,22 @@
 }
 
 .c-article-subject-list {
-	padding: 0;
 	font-size: map-get($article--font-sizes, 'font-size-2');
 	font-family: $font-family-sans;
+	list-style: none;
+	display: flex;
```
saramorueta: is it necessary to add here a fallback for IE11?, I think `flex-wrap` is not compatible. Have you double checked?

moddular: Oh, I didn't check - I just copied what was in BMC - I'll take a look!

moddular: Checked an article where the keywords wrap onto multiple lines and it looks ok in IE 11 - so I don't think we run into any of their flex-wrap bugs

PR #174, opened 2019-06-26
Update share-box js to allow all imprints to use same code

nickcall:
This update allows the passing of the readcube API endpoint from each environment. As BMC and Nature use different endpoints currently.

It also adds a style for any information that may appear if the request sharedIt link is not available.


packages/global-article/HISTORY.md
```diff
@@ -1,5 +1,9 @@
 # History
 
+## 2.0.0 (2019-06-26)
+	* Update share-box.js so readcude api endpoint can be passed in allowing all environemts to use same code
```
saramorueta: *environments*

packages/global-article/js/share-box.js
```diff
@@ -3,19 +3,22 @@
  *
  */
 
-window.Component.ShareBox = (function (doc) {
+var ShareBox = function () {
 	'use strict';
+	var doc = document;
```
saramorueta: I'd suggest replacing the actual variable for its value in all uses. I think there are only 6 uses :)

packages/global-article/js/share-box.js
```diff
@@ -195,8 +204,12 @@ window.Component.ShareBox = (function (doc) {
 			var doi = (config && config.doi) ? config.doi : null;
```
saramorueta: I'd add validation for the `url` as well as it's done with the `doi` and avoid calling to `_addGetShareUrlClickEvent ` if it's null.

packages/global-article/HISTORY.md
nickcall: done

packages/global-article/js/share-box.js
nickcall: done

packages/global-article/js/share-box.js
nickcall: have added the appropriate code

PR #175, opened 2019-06-27
Add unit test to js components

saramorueta:
I'd like to merge what I've done so far here despite it's not 100% completed. So that we can start adding more tests gradually.

As part of the transition to ES6 and in order to be able to test the code, I had to export each file individually by replacing the IIFEs with named functions to be exported separated. Thus this branch is going to add a breaking change in the history.

Screen Shot 2019-09-23 at 14 00 09


benjclark:
Nice work. Out of interest, for the modules that you have converted from IIFE to newer syntax - do they need to be imported by the consumer (i.e. oscar-sites-nature js files) in a different way or does that stay the same?

saramorueta:
Unfortunately, we need to adapt the way that use of these components a bit. The current implementation of these components is returning in most of the cases an `object` with some public methods, so for now we would need to stick to that which means that we'll have to create a new instance of these objects before being able to use them.

on the shared article renderer world, we define a namespace called `Component` so it's easier for now just to do this:

```
Component.ComponentName = new Component();
Component.ComponentName.init();
```

saramorueta:
@benjclark I think this is good to merge after my latest changes. I have tested those in the three oscar-sites and seem to work fine. 

Note: Before using this version it's needed to do some minor adaptation in the three oscar-sites to load the components. I have these changes locally, just ready to push once this version is deployed,

moddular:
This looks wonderful 😍 I think you can delete collapsible-sections.js and it's tests though - it doesn't exist anymore

saramorueta:
Ooh! that's very true! thanks for that @moddular

PR #178, opened 2019-07-02
Create global-context JS helpers

davidpauljunior:
- Creates a JS folder in `global-context`
- Adds helper folder, split into `util.js` and `dom.js`
- `dom.js` is purposefully blank, it's a placeholder
- Adds `makeArray` helper and test


alexkilgour:
Nice. Only comment, can we link to the helpers `README.md` from the main readme - stick an item in the contents list at the top that links down to a section where you explain that we have js helpers and link off to the page

PR #179, opened 2019-07-02
Custom borders padding

saramorueta:
I'm not sure if this component is deprecated but it fits our needs right now, so I've decided to add some minor modifications to make it a bit more versatile.

Modify box mixin to allow receive multiple border positions and paddings.


packages/global-box/scss/30-mixins/_box.scss
```diff
@@ -6,10 +6,31 @@
 @mixin box (
 	$padding: $box--padding,
 	$border: $box--border-color,
+	$border-directions: (top, right, bottom, left),
 	$background: $box--background-color
```
alexkilgour: This should be set in the settings file as `$box—border-directions` than mapped here to `$border-directions` as with the other settings

alexkilgour:
This was meant to be an example component as it is very simple, and also quite limited - you set everything in the settings then every box has to look the same. In my opinion, if this is useful it would be better served existing as a component within each application that uses it.

saramorueta:
Ok. I see your point. It was just me playing with the idea of doing something more useful but you're right and this is too much simple to even exist. I should have seen on my own.

I'd say to clean the repo a bit to avoid doing things like this one in the future.

saramorueta:
This doesn't apply anymore.

PR #180, opened 2019-07-02
Add global banner

saramorueta:
This has to be merged after https://github.com/springernature/frontend-global-toolkit/pull/179/

Screen Shot 2019-07-02 at 17 13 03

Screen Shot 2019-07-02 at 17 16 30


packages/global-banner/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@springernature/global-banner",
+  "version": "0.0.1",
+  "license": "MIT",
+  "description": "Represents a box with text",
+  "keywords": [],
+  "author": "Springer Nature",
+  "peerDependencies": {
+    "@springernature/global-box": "3.1.0"
+  }
```
alexkilgour: We do not allow inter-dependent components. The only dependency a component should have is the context (and every component should have that as a dependency). Just hardcode the box styles into this component.

packages/global-banner/scss/10-settings/_banner.scss
```diff
@@ -0,0 +1,18 @@
...
+$banner--padding-x: 36px;
+$banner--border-color: #25426f;
+$banner--background-color: #dceaf6;
+$banner--font-family: 'serif';
+$banner--font-size: 14px;
+$banner--line-height: 1.4;
+$banner--link-color: #0067c5;
+$banner--link-visited: #8e2555; 
+$banner--link-hover: #0097ef;
+
```
alexkilgour: Can we make the settings here more generic - colors should be values like `black`, `red` rather than ‘branded’ colours, and the same with the padding values.

packages/global-banner/scss/30-mixins/_banner.scss
```diff
@@ -0,0 +1,40 @@
...
+		color: map-get($link, 'color');
+
+		&:visited {
+			color: map-get($link, 'visited');
+		}
+
+		&:hover {
+			color: map-get($link, 'hover');
+		}
+	}
```
alexkilgour: Do the link styles not just inherit from the context? why would you want different link styles within the banner to the rest of the application?

alexkilgour: The same with something like `font-family`. If a banner should have serif text then you can reference `$font-family-serif` directly from the context in your variable

saramorueta: well.. it makes sense to me if you have the option of customising the `background-colour` although I think we'll rarely use a background that needs a different colour of links than the rest of the application. I'll remove it then.

regarding the font-family, you're right, I didn't add it initially because I wasn't sure about if I could add the context as a dependency at this level.

saramorueta:
After discussing with @ricky-lau, we decided to add only a class called`c-banner` as default and add variations (css modifiers) of it to override specific properties. It'd be clearer to read and more clear and simple to use.

ricky-lau:
Made changes to this PR to accommodate customer branding banner along with existing Nature feedback banner. 

Added new examples in readme about how to use the banner in relation to layout. Layout classes `u-container` already exist at the product level for SpringerLink and BMC/SO. Nature requires a combination of `content mq1200-padded` to get equivalent styles.

PR #182, opened 2019-07-03
Prevent section titles containing links from breaking the reading com…

moddular:
…panion section navigation


packages/global-article/js/reading-companion.js
```diff
@@ -27,8 +27,19 @@ window.Component.ReadingCompanion = (function (win, doc) {
 
 	var _container = null;
 
+	function cleanLinks(html) {
+		var d = doc.createElement('div');
+		d.innerHTML = html;
+		d.querySelectorAll('a').forEach(function(a) {
```
saramorueta: Even if we add a polyfill to cover this iteration over the _nodelist_, I don't think it's good to use it extensively because we could create a false idea of compatibility. I'd stick to use the more safe syntax `Array.prototype.slice.call(document.querySelectorAll('a'));` despite it's longer. Obvs, this is my personal preference ;)

PR #184, opened 2019-07-04
Fix PDF download link hover styles for BMC

moddular:


moddular:
The PDF link has a dark blue background and BMC links have a light blue hover colour which ends up having really bad colour contrast. Nature links don't change colour on hover so I didn't notice there. We can probably do something nicer like having the link background change colour - but for now this matches how the link looks on Nature so I thought it would be enough to at least ship it :)

PR #194, opened 2019-07-22
Transform block of references to use flex

saramorueta:
There was a set of properties that weren't working well in all cases. So I've decided to use `flex`


jpw:
Does this syntax work in IE10? (sorry)

saramorueta:
thanks for reviewing this! 
That's a very good question indeed! Nothing to worry about though as the auto prefix plugin does the magic :)
I added a small margin-left to the list that was missed!

PR #197, opened 2019-07-24
Update utility class to add heading styles

nickcall:
Currently adds font-size information for `<h1>` to `<h4>` 

Can be used by adding style `u-h1` etc to element.


packages/global-context/scss/60-utilities/typography.scss
```diff
@@ -60,3 +60,23 @@
 	font-weight: normal;
 	font-style: normal;
 }
+
```
morgaan: Hi,

Here is a suggestion:

Since all it does is to set the proper font-size, I would make the size a variable that could be overridden in the settings layer.

Rather than naming it after the type of heading I would name it after the text size (eg. `.u-text-huge` or `.u-text-xxl`).
The font-size property can then be generated via a mixin that takes care of providing both `rem` value and its `px` fallback. The mixin could take the corresponding text size variable as parameter. The font-size rule can then be implemented the same way for the `h1`, `h2`... elements.

I believe this enhance the re-usability of the class and is more bound to what the utility class actually does.

packages/global-context/scss/60-utilities/typography.scss
```diff
@@ -60,3 +60,23 @@
...
+	font-size: 2.4rem;
+}
+
+.u-h3 {
+	font-size: 17px;
+	font-size: 1.7rem;
+}
+
+.u-h4 {
+	font-size: 14px;
```
morgaan: For accessibility reason I would recommend to not go below 16px as much as possible

packages/global-context/scss/60-utilities/typography.scss
nickcall: Hi Morgan,

I agree with what your saying, I will take a look at a soon at all font-sizes and maybe extract the sizing out into variables which could then be overridden, this is the first cut to use in the shared renderer for articles, I will next need to look at the base font used in the `HTML` and `body` attributes at that point I will look at creating variables for sizing.

Thanks,
Nick

packages/global-context/scss/60-utilities/typography.scss
nickcall: That just replicates what we have at the moment

PR #205, opened 2019-07-30
Remove unused flex mixins

ricky-lau:
- remove `u-flex-split` and `u-flex-unsplit`
- bump global-context to version 3.0.0


davidpauljunior:
Couple of things.
You need to add an entry into the HISTORY.md.
Also this is a breaking change as it's removing code that people may have.
Even if our repos don't use those classes, in theory someone could install an version of global context, and they would automatically get this as it's a patch.

PR #208, opened 2019-08-02
Update autosave

jon-stevens:
This update allows the consumer to optionally provide data to the component rather than have it retrieve it from an endpoint.


packages/global-autocomplete/README.md
```diff
@@ -39,7 +39,8 @@ const args = {
     selector: '[data-component-autocomplete]',
     onSelect: onSelect,
     searchError: onError,
-    endPoint: 'autocomplete?q=',
+    endPoint: 'autocomplete?q=',  // OPTIONAL: Provide an endpoint if you want the component to request data
```
jpw: `endpoint`

packages/global-autocomplete/js/index.js
```diff
@@ -1,21 +1,34 @@
...
+		minChars = 0,
+		componentName,
+		data,
+		onSelect,
+		inputDelay = 300,
+		requestTimeout = 2000,
+		headers = {},
+		searchError,
+		resultsCallBack
+	} = args;
```
jpw: ohhhh that's a nice pattern, i will have to steal that 🙂

packages/global-autocomplete/js/index.js
```diff
@@ -1,21 +1,34 @@
...
+	const input = document.querySelector(selector);
+
+	if (!endPoint && !data) {
+		console.error('Please provide data or a data endpoint');
+		return;
+	}
+
 	const container = () => {
-		return document.querySelector(`.${args.resultsContainerSelector}`);
+		return document.querySelector(`.${resultsContainerSelector}`);
```
jpw: Should it be a restriction that `resultsContainerSelector` has to be a class? Docs imply not...

packages/global-autocomplete/js/index.js
```diff
@@ -1,21 +1,34 @@
...
 	};
 	const suggestions = () => {
-		return Array.from(document.querySelectorAll(`.${args.resultSelector}`));
+		return Array.from(document.querySelectorAll(`.${resultSelector}`));
```
jpw: ditto

packages/global-autocomplete/js/index.js
```diff
@@ -1,21 +1,34 @@
...
+		data,
+		onSelect,
+		inputDelay = 300,
+		requestTimeout = 2000,
+		headers = {},
+		searchError,
+		resultsCallBack
+	} = args;
+
+	const input = document.querySelector(selector);
```
jpw: Is `selector` a required arg?

packages/global-autocomplete/js/index.js
```diff
@@ -122,9 +135,14 @@ const autoComplete = args => {
 		}
 	};
 
-	const sendQuery = term => {
-		let getSuggestions = new Promise((resolve, reject) => {
-			fetch(endpoint + term, {
+	const handleData = term => {
+		if (data) {
```
jpw: A pet peeve of mine is `data` as a variable name. 

1. By the time I got to this point in the code, I had already forgotten what `data` referred to.
1. In `generateSuggestions` above, you also use `data` locally (which is fine there IMO as it's dealing with fetched _and_ supplied result data, so it's agnosticism as to the source of the data makes sense), whereas here you are referring to the `data` argument... Maybe a more descriptive name for the argument? `resultData`? `resultDataArg`? `staticResultData`? `suppliedResultData`?

packages/global-autocomplete/README.md
```diff
@@ -39,7 +39,8 @@ const args = {
...
+    data:                   // OPTIONAL: Provide JSON if you want to fetch data and transform it yourself.
```
jpw: If I am reading the code correctly, `data` takes precedence over `endpoint` if both are supplied..? maybe that should be mentioned here?

jon-stevens: ah yes, good point

packages/global-autocomplete/js/index.js
jon-stevens: yeah good point

packages/global-autocomplete/js/index.js
jon-stevens: Yep. I have now added a check

packages/global-autocomplete/js/index.js
jon-stevens: 👍

PR #210, opened 2019-08-05
Display fixes

moddular:


moddular:
11px! What were we thinking!? :)

PR #215, opened 2019-08-09
Remove abbr from open access label

moddular:


packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -2306,9 +2306,8 @@
 		border-right: 0;
 	}
 
-	abbr {
+	&--open {
```
saramorueta: humm.. that modifier looks weird to me. IMHO that `abbr` is an element. If the html structure corresponds to this:
```
                            <li class="c-article-identifiers__item">
                                <abbr title="Open Access">Open Access</abbr>
                            </li>
```
Why do you consider it as a modifier of its parent?

moddular: I'm proposing the html becomes:

```
 <li class="c-article-identifiers__item c-article-identifiers__item--open">
    Open Access
</li>```

since the text in `abbr` element isn't an abbreviation of anything anymore

saramorueta: oh, I've just noticed that the use of `abbr` here is incorrect.

PR #218, opened 2019-08-12
Add c-article-header class to align the article header - title, autho…

moddular:
Add c-article-header class to align the article header - title, author list etc - with the body sections


packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -146,6 +146,10 @@
 	clear: both;
 }
 
+.c-article-header {
```
saramorueta: I think this one is tricky in regards the different breakpoints. Have you checked this change is behaving as expected in all cases?

moddular: Good catch - thanks!

moddular: Updated to take this into account

PR #219, opened 2019-08-13
Remove bootstrap from global-article.

saramorueta:
Basically this file manages the nature header and handle how other elements in the page relates to it. It seems to be quite specific for nature, so I decided to take it out from here.

I'll include it separately in nature.


packages/global-article/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 4.6.0 (2019-08-13)
```
moddular: The version number doesn't match up with the one in package.json

saramorueta: oh, nooo! sorry

PR #222, opened 2019-08-21
Remove jQuery from popup and simplify to remove functionality not nee…

moddular:
…ded by the author popups which are the only remaining usage


moddular:
Thanks Sara :)

PR #227, opened 2019-09-13
Remove margin-left rule on list-reset

davidpauljunior:
`li`s don't have a margin left by default, so there's no need to reset it.
Because the `li` is a nest selector, it is trumping utilities.


benjclark:
breaking change I think? as anyone who relied on that rule will now potentially have broken styles

davidpauljunior:
yeah maybe

PR #230, opened 2019-09-17
Updates styles to move article metrics into body

nickcall:
Move the article metric bar that is currently above the reading companion to below the article title area in the main body. This has been done to give more space for future RA21 developments.


nickcall:
After feedback from @hollsk I have removed the use of `<dfn>`

PR #233, opened 2019-09-17
add flag to manage visibility of the provided by text

saramorueta:
To better handle the visibility of the whole box for the provided by text under the ra21 flow, I added this flag that allows to display or not the text optionally and keep displaying the rest of the component.


benjclark:
out of interest, what is "provided by box"? what does it do and look like?

saramorueta:
@benjclark I agree with you that the name is everything but good. I'll probably change it later, as now that this component is becoming something bigger in dimension. 
Here another PR with some screenshoot. I'm happy to show you more in detail if you're interested. 
https://github.com/springernature/oscar/pull/1008

@moddular well, part of the change might help with that issue, specifically moving the assignment of the parentElement to the init method, because I noticed that sometimes this is `null`. But I don't trust 100%. There is another change in this PR to basically allows us to control the visibility of the whole box instead of just the "Access provided by insitutionName" avoiding that way a jumping effect when the institution name is added to the html.

PR #236, opened 2019-09-27
Allow ShareBox to accept only a url

moddular:


packages/global-article/js/share-box.js
```diff
@@ -200,11 +198,12 @@ var ShareBox = function () {
...
 		init: function (config) {
-			var doi = (config && config.doi) ? config.doi : null;
-			var url = (config && config.url) ? config.url : null;
+			var doi = (config && config.doi) ? config.doi : '';
+			var url = (config && config.url) ? config.url : '';
+			var source = url + doi;
 
-			if (doi && url && _renderInto(document.querySelector('[data-component="share-box"]'))) {
-				_addGetShareUrlClickEvent(doi, url);
+			if (source && _renderInto(document.querySelector('[data-component="share-box"]'))) {
```
saramorueta: Even if it is very unlikely that happens, an empty string will be truth in this case. I'd add `source !== ''"` to make sure that doesn't happen.

packages/global-article/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 6.0.4 (2019-09-27)
+	* Allow `ShareBox` to accept only a url or a doi in it's config
```
saramorueta: typo here :) `its`

PR #245, opened 2019-10-09
Support gradient to cta buttons

saramorueta:


packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -2135,7 +2135,11 @@
 .c-pdf-download__link {
 	display: block;
 	color: $article--cta-button-color;
-	background: $article--cta-button-background-color;
+	@if variable-exists(article--cta-button-background-gradient-top) {
```
saramorueta: I'm wondering if instead of using `variable-exists` check and have to manage two different variables, if it would be better to have a default background in the variable `$article-cta-button-background-color` and then override its value in the consumer, just setting it to that `linear-gradiant`

PR #248, opened 2019-10-14
Added underline modifier for link element on global banner

ricky-lau:


packages/global-banner/scss/50-components/_banner.scss
```diff
@@ -115,3 +115,13 @@
 		color: $banner--link-color;
 	}
 }
+
+.c-banner__link--underline {
```
saramorueta: I'm wondering if this style should be part of the default one.
James A. shared recently this https://app.zeplin.io/project/5cc8118dc2e0fc1834f9adc4/screen/5d6d3fe4f555509eedd3eaab as reference.

ricky-lau: Yea I agree. I would say when it gets added to the global toolkit then this part can get removed or replaced by it 😬

saramorueta: fair enough!

PR #260, opened 2019-10-16
Add demo folder to global-autocomplete

jpw:
- we want to show working demos of components in the Elements web site, e.g:

image

- doing this requires a `demo/` folder in component source code, which will be processed by a "package renderer" utility I am working on.
- [Here's the idea behind how the demo code works](https://gist.github.com/alexkilgour/be59684689020c2a4aea3d559b360dce) but it basically includes the brand context component, has some entrypoint files, and some sample usage JS.
- this will remove the need for extensive READMEs as developers will be able to look at the demo code.
- my "package renderer" utility at the moment renders this component demo folder a bit this:
image
- my "package renderer" utility will also be used for a local dev server we are planning, which will make it easier to develop components outside of the rest of an app's source code and create the contents of this demo folder.

This PR is a first-pass at such a demo folder, and is based on a mock package I am using to test the package renderer.


jpw:
It's not going to publish for a bit though, as work will need to be done on the package validator before Travis is appeased.

packages/global-autocomplete/demo/main.js
```diff
@@ -0,0 +1,512 @@
...
+	endpoint: 'autocomplete?q=',
+	timeout: 2000,		// OPTIONAL: Set a timeout for the fetch request, onError will be called if fetch request timeouts, default is 2000
+	minCars: 1,			// OPTIONAL: Minimum characters to be typed before request is sent, default is 0
+	inputDelay: 300,	// OPTIONAL: Delay between keypress and request being sent, default is 300
+	headers: {
+		Accept: 'application/json; version=2'
+	},
+	resultsContainerSelector: 'c-results-container',
+	resultSelector: 'c-results-container__result',
+	resultsCallBack: showResults
```
jon-stevens:
```suggestion
	staticResultsData: animalsList
```

You could have done this instead of hijacking window.fetch, doesn't make much difference for the purposes of this PR but would be a bit less clutter

jon-stevens:
image

packages/global-autocomplete/demo/main.js
jpw: yep ta, not sure whether to do that or not now, will have a think on it...

jpw:
NICE

PR #272, opened 2019-10-25
Remove download citation and right chevron icons as background images

moddular:


moddular:
> I guess the colour is inherited from their parent, isn't it?
Yeah - there's a u-icon class that sets the fill to be currentColor - so the icon matches the surrounding color - handy for things like hover styles.

PR #285, opened 2019-11-04
Updates global font family sans to system fonts

nickcall:
As per https://github.com/springernature/oscar/issues/1077


benjclark:
should this be a major new version? as (hypothetically) it "breaks" the look of a webpage that relied on the old font being there (lets say for argument sake that a design had been in place, this would break that exact design)

nickcall:
Yeah I was undecided about that, I did feel it was backward compatible but can see what you mean.

PR #287, opened 2019-11-06
update readme to reflect data options

jon-stevens:
Small update to demo to illustrate using static data rather than using an endpoint to retrieve results


packages/global-autocomplete/demo/main.js
```diff
@@ -29,7 +29,8 @@ const args = {
 	selector: '[data-component-autocomplete]',
 	onSelect: onSelect,
 	searchError: onError,
-	endpoint: 'autocomplete?q=',
+	endpoint: 'autocomplete?q=', // OPTIONAL: Provide either endpoint or array of data - if both are provided static data will be used
+	//staticResultsData: animalsList, // OPTIONAL: Provide either endpoint or array of data - if both are provided array be used
```
davidpauljunior: Is the `//staticResultsData` part meant to be commented?

jon-stevens: It's just to illustrate that you don't need to pass in both an endpoint and the static data.

I did want to pass in staticResultsData and comment out endPoint instead but I didn't want to potentially break Jon's demo

davidpauljunior: Oh I get ya, sorry.

packages/global-autocomplete/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 1.0.2 (2019-10-16)
+    * Update readme
```
jpw: The date is wrong, and this PR does not update the readme? :)

packages/global-autocomplete/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 1.0.3 (2020-01-06)
```
alexkilgour: also update package.json file to new version?

PR #295, opened 2019-11-19
Global Expander

davidpauljunior:
Adds a global expander package (click a trigger, toggles a class on a target).


packages/global-expander/README.md
```diff
@@ -0,0 +1,51 @@
...
+``` 
+
+### Options
+
+| Option             | Default Value | Type    |
+|--------------------|---------------|---------|
+| TARGET_HIDE_CLASS  | 'u-hide'      | String  |
+| TRIGGER_OPEN_CLASS | -             | String  |
+| TRIGGER_OPEN_LABEL | -             | String  |
+| CLOSE_ON_CLICKOFF  | false         | Boolean |
```
jon-stevens: Should CLOSE_ON_CLICKOFF be true?

jon-stevens: Does CLOSE_ON_CLICKOFF default to true?

packages/global-expander/js/expander.js
```diff
@@ -0,0 +1,182 @@
...
+
+		do {
+			if (target === document.body) {
+				this._close();
+				this._removeTemporaryEventListeners();
+				return;
+			}
+
+			target = target.parentNode;
+		} while (target !== this._targetEl && target !== this._triggerEl && target !== null && target.nodeType === 1);
```
jon-stevens: Could you do

```
		if (this._targetEl.contains(target) || this._triggerEl.contains(target)) {
                               return;
                }
```

instead of do loop?

packages/global-expander/js/expander.js
```diff
@@ -0,0 +1,182 @@
...
+		let target = e.target;
+
+		if (target === this._targetEl || target === this._triggerEl) {
+			return;
+		}
+
+		do {
+			if (target === document.body) {
+				this._close();
+				this._removeTemporaryEventListeners();
```
jon-stevens: is  this._removeTemporaryEventListeners is already called inside the _close method?

packages/global-expander/README.md
```diff
@@ -0,0 +1,51 @@
...
+const myExpander = new Expander(trigger, target, options);
+
+myExpander.init();
+``` 
+
+### Options
+
+| Option             | Default Value | Type    |
+|--------------------|---------------|---------|
+| TARGET_HIDE_CLASS  | 'u-hide'      | String  |
```
jon-stevens: Should this be u-js-hide

packages/global-expander/js/expander.js
davidpauljunior: NICE! Just added that and added some more tests to cover the clicking off / clicking a child of the target.

davidpauljunior:
Thanks a lot @jon-stevens, that was brilliant 👍 

Have made all your suggested changes 🎉

packages/global-expander/__mocks__/@springernature/global-context/js/helpers.js
```diff
@@ -0,0 +1,44 @@
...
+			if (attributeValue) {
+				dataOptions[key] = attributeValue;
+			}
+		}
+	}
+
+	return dataOptions;
+};
+
+const makeArray = iterable => {
```
jon-stevens: Did you add this helper because Array.from is not supported in ie?
There is a polyfill you could specify in the README. Or I just use something like [].slice.apply()

davidpauljunior: No the reason for the helper is because I found that dotted through our code base was arrays being made by using `slice.apply()`, `Array.from`, and `[...nodeList]` and that made the polyfilling a bit annoying.  I wanted 1 single way to create an array and this was faster than all the others and needed no polyfills.

jpw: This PR warmed my cockles. 
> I wanted 1 single way to create an array and this was faster than all the others and needed no polyfills.

Sounds like a playbook thing maybe..? 😁

PR #301, opened 2019-11-22
Add card component

ricky-lau:
initial commit to getting this into global and then test changes locally

- removed u-faux-block-links and u-position-relative utility classes in html and made it part of .card to make card clickable


packages/global-card/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@springernature/global-card",
+  "version": "0.0.1",
```
davidpauljunior: 0.1.0

packages/global-card/HISTORY.md
```diff
@@ -0,0 +1,4 @@
+# History
+
+## 0.0.1 (2017-11-22)
```
davidpauljunior: 0.1.0

PR #302, opened 2019-11-25
add peerDependencies to global-context on global author-list, banner and card

ricky-lau:
- also correct some dates in history


benjclark:
I thought we were using https://github.com/springernature/frontend-toolkits now?

ricky-lau:
We are, but the global toolkit hasnt been moved over yet

PR #308, opened 2019-12-04
Set cursor position of focused element

davidpauljunior:
Moves the cursor to the end of the focused input value if applicable.
Useful for things like when the first tabbable element inside a target is a search box with a prepopulated value.


ricky-lau:
Not directly related to the changes but I have a feeling that this should focus on the first element in expander rather than first tabbable so that screen readers like VoiceOver doesn't skip over it when opening the expander. I think I came across this on the accessibility report on SpringerLink... I'll have to dig this out.

davidpauljunior:
@hollsk that's much better! I'll do that instead

davidpauljunior:
@ricky-lau that is a very interesting point. That would possibly mean making the target itself have focus, but then you don't get the niceness of entering into the first field.  Maybe that's actually not a niceness. @hollsk ?

davidpauljunior:
@hollsk I made the change so that it now highlights the text.
I'm interesting in what you think about Ricky's comments though.

hollsk:
I don't have enough context to comment! Can I see a working version anywhere? Or a version that demonstrates the old behaviour that you want to change?

hollsk:
Dave has shown me the thing and it is indeed context-dependent 😁 

In the working example that Dave showed me, the first element is the label for the text input, so it's fine as it gets read out anyway when the input is focused (or it _should_ do). 

If there's a possibility that there may be anything else above the text input/label in the component, then I strongly agree with Ricky. Autofocusing any field always carries a risk that screen reader users might miss content that comes before the field, and I usually try to discourage people from doing it. 

In _this specific case_ I think it's fine because the autofocus is taking the user to the place where they expect to carry out the task they intend to complete. If you're pretty sure that people will use the component as-is then I'm ok with it.

ricky-lau:
Here's a screenshot of what I was trying to remember. And as @hollsk already mentioned it is only applicable when there is something above focused input. 

image

davidpauljunior:
@ricky-lau @hollsk prepare to be dazzled.
I made `autofocus` an option that defaults to false. Now you can set it conditionally 🎈

davidpauljunior:
Oh, and just to clarify, if you don't set that option, I add a `tabindex` of `-1` to the target so that the user can tab through the content of the target.

PR #309, opened 2019-12-13
Remove underscore from filename

davidpauljunior:


davidpauljunior:
Opted to change the filename in brands to use an underscore instead.

